PR #13, opened 2021-02-12
Add Vite Rails Legacy, which targets Rails 4

ElMassimo:
[vite_rails]: https://github.com/ElMassimo/vite_ruby/tree/main/vite_rails
[vite_ruby]: https://github.com/ElMassimo/vite_ruby/tree/main/vite_ruby

### Description 📖

This pull request adds `vite_rails_legacy`, which is [__Vite Rails__][vite_rails], modified to support Rails 4.

## Why a separate library? 🤔

I don't want the burden of having to support Rails 4 in the main codebase.

And yet, after extracting [_Vite Ruby_][vite_ruby] the surface area is smaller,
and it was easy to adapt it to work with Rails 4.

This way, people get to use the library in Rails 4, and I get to not worry about
accidentally breaking support with changes that are tested on Rails 5 and onwards.

### Screenshots 📷

Running on Rails 4.2

Screen Shot 2021-02-12 at 12 44 00

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -3,6 +3,7 @@
 [webpacker]: https://github.com/rails/webpacker
 [vite rails]: https://github.com/ElMassimo/vite_ruby
 [vite]: https://vitejs.dev/
+[vite_rails_legacy]: https://github.com/ElMassimo/vite_ruby/tree/main/vite_rails_legacy
 [vite_hanami]: https://github.com/ElMassimo/vite_ruby/tree/main/vite_hanami
 [vite_ruby]: https://github.com/ElMassimo/vite_ruby/tree/main/vite_ruby
 [commands]: /guide/development.html#cli-commands-⌨%EF%B8%8F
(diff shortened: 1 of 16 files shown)


CyberAP:
Thanks for the legacy version of this gem, this is much appreciated! Unfortunately ruby requirements are too high for it to be actually useable (it requires ruby 2.5 and it has been supported only since Rails 5.1). Is there any way to make it work with Ruby 2.2.1 or lower?

ElMassimo:
Unfortunately `2.2` does not support the safe navigation operator (`&.`), which is also used in `vite_ruby`.

You could fork both `vite_rails` and `vite_ruby` and replace usages of the safe navigation operator with conditional checks.

[Support for Ruby 2.2 ended long ago](https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/).

patrickdet:
I am running into the same issue which is Ruby 2.4.10.0 which does support the `&.` operator. Any chance of weakening the Ruby version requirement to `>=2.4`?

ElMassimo:
Hi @patrickdet.

[Support for Ruby 2.4 ended last year](https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/), but I understand that sometimes upgrades are not viable.

I have a proposal for you–if you fork the repo and test both `vite_rails_legacy` and `vite_ruby` with lowered requirements, test the gems, and they work as expected without major changes, we can bring in those changes and I'll release a new version of both gems.

patrickdet:
Hi @ElMassimo, sounds good. I have opened a PR #84 which brings these changes to 2.4.

PR #39, opened 2021-03-19
fix: handle getaddrinfo errors when checking dev server

palkan:
### Description 📖

Fixes `#dev_server_running?` failing with an exception in case hostname is not resolveable.

### Background 📜

That's, for example, the case when using a dockerized development environment (and not running a vite-dev server at all).

### The Fix 🔨

I suggest removing the exception class filter altogether like [Webpacker does](https://github.com/rails/webpacker/blob/da362b8c96a4be5a4af8f8ad886f5dd8451f457f/lib/webpacker/dev_server.rb#L21).

### Screenshots 📷

image


ElMassimo:
Usually a dockerized setup would also start the Vite dev server, but I don't see a reason to be strict about the check.

Thanks for the fix!

PR #51, opened 2021-04-14
Correct non-working links

coorasse:
### Description 📖

Corrects non-working links

### Background 📜

I was clicking on the links and they were not working 😱 

### The Fix 🔨

I use working links which reflect the initial intention


ElMassimo:
Thanks Alessandro!

PR #63, opened 2021-04-22
deps: Update dependencies after `mimemagic` scandal

doabit:
### Description 📖

Update rails version to fix error caused by `mimemagic`.

### Background 📜

`mimemagic` may cause error when run `bundle install`

```
An error occurred while installing mimemagic (0.3.10), and Bundler cannot continue.
Make sure that `gem install mimemagic -v '0.3.10' --source 'https://rubygems.org/'` succeeds before bundling.
```


ElMassimo:
Hi Sean, thanks for the contribution!

Would you modify this PR so that only bumps `mimemagic`?

doabit:
Hi, rails has removed `mimemagic ` and  replaced it with  `mini_mime`.

ElMassimo:
Sounds good, I checked the dependencies and it shouldn't cause issues. Thanks!

PR #67, opened 2021-04-23
fix: Update installation to use the latest version of the plugin

doabit:
### Description 📖

When run `bundle exec vite install`, still use plugin `1.0` version, i think we should use the latest version.


doabit:
Sorry, i still can't find out why it still failed.

ElMassimo:
Thanks for the reminder!

PR #110, opened 2021-07-29
fix: use asset host if set during development

StefSchenkelaars:
### Description 📖

This pull request fixes the loading of the Vite client and react refresh preamble when an `asset_host` is defined and the dev server is running.

### Background 📜

This was happening because when an asset host is defined and a dev server is running the vite client is loaded twice. Once with the `asset_host` domain and once with the relative path (so the current domain).

Consider the case where all assets are provided by `my-cdn.com`. The Vite client loaded with the `vite_client_tag` tag will be loaded without any subdomain (`/vite-dev/@vite/client`) while the client as fetched from within the code will be loaded from the same domain as the asset itself (in this case `my-cdn.com/vite-dev/@vite/client`).

This results in two vite clients being loaded and a crashing page.

### The Fix 🔨

By changing the prefixing method of the vite assets with the assets subdomain.

### DIFF
--- a/test/manifest_test.rb
+++ b/test/manifest_test.rb
@@ -93,6 +93,14 @@ class ManifestTest < ViteRuby::Test
     }
   end
 
+  def test_lookup_success_with_asset_host_and_dev_server
+    entry = { 'file' => 'http://example.com/vite-production/application.js' }
+    refresh_config(asset_host: 'http://example.com')
+    with_dev_server_running {
+      assert_equal entry, lookup!('application.js')
+    }
+  end
+
   def test_lookup_nil
     assert_nil lookup('foo.js')
   end
--- a/vite_ruby/lib/vite_ruby/manifest.rb
+++ b/vite_ruby/lib/vite_ruby/manifest.rb
@@ -122,7 +122,7 @@ private
 
   # Internal: Scopes an asset to the output folder in public, as a path.
   def prefix_vite_asset(path)
-    File.join("/#{ config.public_output_dir }", path)
+    File.join(config.asset_host || '/', config.public_output_dir, path)
   end
 
   # Internal: Resolves the paths that reference a manifest entry.


ElMassimo:
Thanks for your contribution Stef, looks good!

Out of curiosity, would you explain your current use case for setting `asset_host` during development?

StefSchenkelaars:
We have an application that uses subdomains to switch tenants so to mimic this behaviour locally we use the `*.lvh.me` domains to resolve localhost.

Next to this we use a separate subdomain for our assets. This way we can easily put a CDN in front of it. So to ensure this works properly we set our `asset_host` to `http://assets.lvh.me:3000` during development. This way we can for example check CORS settings etc.

StefSchenkelaars:
Thanks for the quick replies and the release 👍

ElMassimo:
Unfortunately the proposed solution causes the lookup to fail in production, because it affects the name of the entries.

I yanked 1.2.19, and will temporarily revert main. I can revisit the implementation tomorrow, another PR is welcome.

StefSchenkelaars:
Ow, for me it seemed to work 😱 I will have a look tomorrow as well! Thanks

StefSchenkelaars:
I had a short look into this and I wanted to check in the ideas with you before trying to fix it and break it again. 
For me the lookup actually seems to work and since I'm kinda new to vite, I'm not really sure why this is failing. So if you have any better ideas, please.

## Path 1: Fix it in vite_rails
The loading of the vite_client can quite easily be corrected in the [ViteRails::TagHelpers#vite_client_tag](https://github.com/ElMassimo/vite_ruby/blob/main/vite_rails/lib/vite_rails/tag_helpers.rb#L9) by converting
```ruby
tag.script(src: src, type: 'module')
```
into 
```ruby
javascript_include_tag(src, type: 'module', extname: false)
```

So that's all good but then the react_refresh_preamble is more problematic. Since it path is generated within `ViteRuby::Manifest`. So it might be an option to explicitly prefix it only for the preamble.

## Path 2: Separate method in vite_ruby
Both the `@vite/client` and `@react-refresh` are no lookups in the manifest and rather the serving of a package directly. So it might be an idea to create a second method next to `prefix_vite_asset` called `prefix_node_asset` to return the full url.

ElMassimo:
@StefSchenkelaars Path 2 has the benefit of fixing the issue for all frameworks that support an `asset_host` configuration.

I've released `vite_ruby@1.2.20`, please upgrade and let me know how it goes!

StefSchenkelaars:
@ElMassimo I've tested it and it seems to be all good 👍  Thanks!

PR #117, opened 2021-08-16
feat: conditionally enhance assets:precompile

dflynn15:
### Description 📖

Add the ability to conditionally run `vite:build` in the `assets:precompile` Rake task.

### Background 📜

I am iteratively migrating my application from using Webpacker to Vite Ruby. While I am doing this I would like the option to temporarily turn off the hook for `vite:build` to be run on `assets:precompile`. By adding this `ENV` check, we introduce an option when running `bin/rails assets:precompile` to opt-out of the enhancement when desired without trying to filter out the `vite:build` proc added as an action.

This is definitely a "niche" use-case for a migration path, but there is a small amount of merit in my mind to providing an opt-out hook on `assets:precompile`.

I chose not to include any version bumps with this PR in case you wanted to manage the release and version changing. I'm happy to include that in the PR as well!

Additionally, I made the PR base from `main` because the only other option was `v3`. If possible, this would be convenient to be released on the `v2` tag but that would require another branch and more work from you 😞. I'm fine either way, but if there is anything I can do to help let me know!

### DIFF
--- a/docs/guide/rails.md
+++ b/docs/guide/rails.md
@@ -108,3 +108,7 @@ When running the development server, these tags are omitted, as Vite will load t
 ```erb
 <script src="/vite/assets/application.js" type="module" crossorigin="anonymous"/>
 ```
+
+### Conditional Compilation 📦
+
+If you want to disable the Rake hook for `assets:precompile` as you integrate `vite_rails` into your application, set the environment variable `SKIP_VITE_COMPILE` to `true`. This will keep your current `assets:precompile` step working without `vite_rails` automagically attaching itself as an action.
--- a/docs/package.json
+++ b/docs/package.json
@@ -11,7 +11,7 @@
     "css": "stylelint --ignore-pattern 'dist' '.vitepress/**/*.(vue|scss|css|postcss)'"
   },
   "dependencies": {
-    "vue": "^3.0.7"
+    "vue": "^3.2.0"
   },
   "devDependencies": {
     "@docsearch/css": "^1.0.0-alpha.28",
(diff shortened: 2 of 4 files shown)


ElMassimo:
Hi Daniel, thanks for writing a detailed description.

In that scenario that you are describing, would it be viable to do this instead:

```ruby
# Gemfile

gem 'vite_ruby', require: false
```

```ruby
# initializer or Rakefile

require 'vite_rails' unless ENV['SKIP_VITE'] == 'true'
```

Could you expand on whether you are running the `vite:build` task manually in the process, or not at all?

dflynn15:
Currently, I am using `require: false` and that does prevent the enhancement from being auto-injected. However, the convenience to have the `vite:build` automagically tied to `assets:precompile` felt slightly overbearing without an escape hatch.

Because it is a unique, niche case and I'm fine if you kick this back. But, here are some more details to help tease out why I decided to present this as an option.

We are using the Strategy pattern to swap between Webpacker and Vite Ruby in a way that a helper was created to `if/else` between the two tags. For completeness, I have tests ensuring the swapping works which I have to have the gem required. I can work around this with `group :development, :test` or adding more changes to our Docker build setup to conditionalize this in a way that injects it in test and other environments that are not "Docker."

Because there can be other use cases or approaches to migrating over I wanted to present a case where the dependency provided an opt-out experience rather than my codebase trying to `require` or add initializers for specific build cases. 

Is this quite a bit self-serving? Yep. I'll totally own that. 

Also, pardon the `vitepress` commits I can yank those out but the docs are not building and that seems important to get working 😂

dflynn15:
@ElMassimo I'm guessing this can be closed since a near-identical change was added here https://github.com/ElMassimo/vite_ruby/pull/116/commits/ae230c4367494d346bc2a51bd20a5beacaa22793 ?

ElMassimo:
@dflynn15 [This feature](https://vite-ruby.netlify.app/guide/deployment.html#disabling-extension-of-the-assets-precompile-task) is now released under `vite_ruby-3.0.0`.

PR #157, opened 2021-12-17
fix: handle new virtual legacy-polyfill name

### LINKED ISSUE #156: Legacy plugin not compatible with vite 2.7.x
[troubleshooting section]: https://vite-ruby.netlify.app/guide/troubleshooting.html

- [x] I have tried upgrading by running `bundle update vite_ruby`.
- [x] I have read the __[troubleshooting section]__ before opening an issue.

### Description 📖
In vite 2.7, the name of the legacy-polyfills entry has changed from `"../../vite/legacy-polyfills"` to `"../../\u0000vite/legacy-polyfills"`. This filename has changed since this polyfill is a so called virtual file ([as explained in the plugin api docs](https://github.com/vitejs/vite/blob/e19ba6a98d85586d1f54e6b0eed11dcf72e6c651/docs/guide/api-plugin.md#conventions)).

This however makes the ruby code crash since it can't resolve the filename (see stacktrace below).

This change is also discussed in https://github.com/vitejs/vite/issues/6097 but as I dove into it I found that this NULL character is not a typo but a feature. So therefore I think the gem should be change to handle this correctly.

If this could be confirmed as a bug, I could work out a solution.

### Reproduction 🐞
- Repo with vite > 2.7 and the legacy plugin
- Render a page with a `vite_legacy_javascript_tag` or `vite_legacy_typescript_tag` tag

### Logs 📜

```
ActionView::Template::Error (path name contains null byte):
vite_ruby (3.0.3) lib/vite_ruby/manifest.rb:176:in `extname'
vite_ruby (3.0.3) lib/vite_ruby/manifest.rb:176:in `with_file_extension'
vite_ruby (3.0.3) lib/vite_ruby/manifest.rb:150:in `resolve_entry_name'
vite_ruby (3.0.3) lib/vite_ruby/manifest.rb:83:in `lookup'
vite_ruby (3.0.3) lib/vite_ruby/manifest.rb:70:in `lookup!'
vite_ruby (3.0.3) lib/vite_ruby/manifest.rb:22:in `path_for'
vite_rails (3.0.2) lib/vite_rails/tag_helpers.rb:22:in `vite_asset_path'
vite_plugin_legacy (3.0.0) lib/vite_plugin_legacy/tag_helpers.rb:30:in `vite_legacy_polyfill_tag'
vite_plugin_legacy (3.0.0) lib/vite_plugin_legacy/tag_helpers.rb:15:in `vite_legacy_javascript_tag'
vite_plugin_legacy (3.0.0) lib/vite_plugin_legacy/tag_helpers.rb:20:in `vite_legacy_typescript_tag'

My app => app/views/layouts/login.html.haml:31
```

### Screenshots 📷

_Provide console or browser screenshots of the problem_.

StefSchenkelaars:
### Description 📖

This pull request ensures that the new polyfill name with the `\u0000` character is handled as expected.

### Background 📜

This was happening because the entry of the legacy-polyfill name has changed to start with a `\u0000` character to mark it as a virtual entry in vite 2.7.

### The Fix 🔨

By changing the lookup to not try to parse the entry name, the `\u0000` character is ignored.

Closes #156

### DIFF
--- a/vite_ruby/lib/vite_ruby/manifest.rb
+++ b/vite_ruby/lib/vite_ruby/manifest.rb
@@ -147,9 +147,11 @@ private
 
   # Internal: Resolves the manifest entry name for the specified resource.
   def resolve_entry_name(name, type: nil)
-    name = with_file_extension(name.to_s, type)
+    unless name.include?('legacy-polyfills')
+      name = with_file_extension(name.to_s, type)
 
-    raise ArgumentError, "Asset names can not be relative. Found: #{ name }" if name.start_with?('.') && !name.include?('legacy-polyfills')
+      raise ArgumentError, "Asset names can not be relative. Found: #{ name }" if name.start_with?('.')
+    end
 
     # Explicit path, relative to the source_code_dir.
     name.sub(%r{^~/(.+)$}) { return Regexp.last_match(1) }
--- a/test/test_app/public/vite-production/manifest.json
+++ b/test/test_app/public/vite-production/manifest.json
@@ -41,7 +41,7 @@
   "_log.d31acc25-legacy.js": {
     "file": "assets/log.d31acc25-legacy.js"
   },
-  "../../vite/legacy-polyfills": {
+  "../../\u0000vite/legacy-polyfills": {
     "file": "assets/polyfills-legacy.07477394.js",
     "src": "../../vite/legacy-polyfills",
     "isEntry": true


ElMassimo:
Hi Stef, thanks for providing a detailed information, a potential fix, and updating the tests 😃

vite_ruby/lib/vite_ruby/manifest.rb
```diff
@@ -147,9 +147,11 @@ def resolve_references(manifest)
 
   # Internal: Resolves the manifest entry name for the specified resource.
   def resolve_entry_name(name, type: nil)
-    name = with_file_extension(name.to_s, type)
+    unless name.include?('legacy-polyfills')
```
ElMassimo, line 150: Perhaps we could take this opportunity to generalize the check:

```ruby
unless name.include?("
```

StefSchenkelaars: I think that a part of your comment didn't come through properly but I think you want to check on the `null` character itself? I think that could work as well. Another solution is to only do this file extension check if a type is explicitly passed. 

But since I wasn't too sure about other effect, I took the minimal change to make it happen. Last time I made a PR, I broke the workflow of other users 🙈 

And maybe also something to take in consideration: If the users using this gem did NOT upgrade vite itself yet. Then this `null` character is not yet in the manifest.

ElMassimo: Sorry, on my phone, hence the conciseness 😅

What I suggest is changing the condition to check for the presence of the null char in the provided name, skipping the file extension check (since adding the extension is not relevant when looking up virtual entries).

It's true that it wouldn't be compatible with Vite 2.6, but we could bump the recommended Vite version to 2.7, which is [checked on startup](https://github.com/ElMassimo/vite_ruby/blob/0850af2170077bd55d54a725fa6030a32bbbd401/vite_ruby/lib/vite_ruby/config.rb#L86), and would be caught during development or in the CI.

ElMassimo:
Thanks for sharing your thoughts about backwards compatibility.

Although I'd like to remove the explicit check for legacy polyfills, I can always do that in later versions, in a major or minor release.

Thanks again for the detailed report and for providing a fix ❤️

StefSchenkelaars:
Yea agree that it would be cleaner to have the check removed. But I also feel that this part is a bit under-tested. I also had the idea of adding the vite version to the test matrix so this could have been catched but that was a bit too much for now

PR #165, opened 2021-12-24
docs: remove sprockets from the Rails app (close #160)

### LINKED ISSUE #160: Document how to remove sprockets (optional)
Edit: Im referring here exclusively to `vite_rails` gem.

Just found out about vite.js and it seems to fill the gap thats missing in all the new paradigm shift going on with rails 7, so I want to give it a shot with a project.

From what I have seen, vite seems to fully replace and the old assets pipeline powered by sprockets, however at the moment of running `bundle exec vite install`, the setup script leaves it coexisting with sprockets. While this may be a a safe and wise default to use, I think most of us would like to completely replace sprockets by vite.js, so some king of flag for the task, maybe like `--rm-sprockets` that would do all the required changes for freeing up a project from sprockets, while obviously leaving vite as the new default asset pipeline would be nice!

Another alternative would be doing it by hand, while its more cumberstone, I think just a quick guide in the docs would do it!

ElMassimo:
### Description 📖

This pull request demonstrates how to remove sprockets from a Rails app, and use Vite Ruby alone for asset management.

For a step by step discussion, refer to [this blog post](https://andre.arko.net/2020/07/09/rails-6-with-webpack-in-appassets-and-no-sprockets/) which discusses how to remove sprockets, and change [`sourceCodeDir`](https://vite-ruby.netlify.app/config/#sourcecodedir) instead of `source_path`.

### Background 📜

In practice, it's perfectly fine to use sprockets and Vite Ruby at the same time, but some users prefer to remove sprockets.


noproblemwiththat:
The blog post above is a dead link, but the Internet Archive has a [copy](https://web.archive.org/web/20250119155754/https://andre.arko.net/2020/07/09/rails-6-with-webpack-in-appassets-and-no-sprockets/)

PR #169, opened 2022-01-01
Add style-src to suggestion Content Security Policy changes

KieranP:
### Description 📖

Add style-src suggestion to content security policy so JS injected styles work in development

### Background 📜

Without this, styles injected via JS into a <style type="text/css"> tag won't work

### The Fix 🔨

By adding style-src, styles added this way in development


ElMassimo:
Looks great, thanks!

KieranP:
@ElMassimo And thank you. I had been looking into integrating Vite on a project of mine to replace Webpack, and your gem worked perfectly without much tweaking. Thank you for putting this gem together.

PR #174, opened 2022-01-12
fix(vite-plugin-ruby): regression in assetHosts protocol

StefSchenkelaars:
### Description 📖

Currently if the asset hosts contains a protocol (https://my-cdn.org for example),
the path.join method from node will replace the double forward slash with a single
one. Therefore the final base URL will become something like:

```
  https:/my-cdn.org/vite
```

This gives problems if vite has to resolve URLs itself. For example if you have:

```
  body {
    background-image: url('../background.png')
  }
```

will resolve in:

```
  body { background-image: url('https:/my-cdn.org/vite/background.png') }
```

which in its turn will call this as a relative path.

### Background 📜

See above

### The Fix 🔨

The fix removes the path.join method with a simple interpolated string.

### DIFF
--- a/vite-plugin-ruby/src/config.ts
+++ b/vite-plugin-ruby/src/config.ts
@@ -101,7 +101,9 @@ function coerceConfigurationValues (config: ResolvedConfig, projectRoot: string,
   // Add the asset host to enable usage of a CDN.
   const assetHost = config.assetHost || ''
   const assetHostWithProtocol = assetHost && !assetHost.startsWith('http') ? `//${assetHost}` : assetHost
-  const base = slash(join(assetHostWithProtocol || config.base || '/', config.publicOutputDir, '/'))
+  const host = assetHostWithProtocol || config.base || ''
+  const suffix = config.publicOutputDir ? `${slash(config.publicOutputDir)}/` : ''
+  const base = `${host}/${suffix}`
 
   const entrypoints = resolveEntrypointFiles(projectRoot, root, config)
   return { ...config, root, outDir, base, entrypoints }


vite-plugin-ruby/src/config.ts
```diff
@@ -101,7 +101,8 @@ function coerceConfigurationValues (config: ResolvedConfig, projectRoot: string,
   // Add the asset host to enable usage of a CDN.
   const assetHost = config.assetHost || ''
   const assetHostWithProtocol = assetHost && !assetHost.startsWith('http') ? `//${assetHost}` : assetHost
-  const base = slash(join(assetHostWithProtocol || config.base || '/', config.publicOutputDir, '/'))
+  const host = assetHostWithProtocol || config.base || ''
+  const base = slash(`${host}/${config.publicOutputDir}/`)
```
ElMassimo, line 105: This would result in a double ending slash if `publicOutputDir` is empty, for which support was added in https://github.com/ElMassimo/vite_ruby/pull/164/files#diff-3fb15ba9c8731f617d892e844af45f68b9f4f904fa4ca07c7e45852c3d1ac5bfR104 (and is the cause of this regression for asset host with a protocol).

To support both use cases, we could do:

```suggestion
  const suffix = config.publicOutputDir ? `${slash(config.publicOutputDir)}/` : ''
  const base = `${host}/${suffix}`
```

At this point it would also make sense to extract lines 102-106 to a helper function and unit test it thoroughly with all of the supported config options, to prevent future regressions. I can take care of that after merging this one.

ElMassimo (changes requested):
Hi again Stef! Thanks for reporting, this is a regression introduced in #164.

vite-plugin-ruby/src/config.ts
StefSchenkelaars: I also tried to write a test case for this but I couldn't find a way of doing that. But I'll update the code to take the double trailing slash into account

StefSchenkelaars: Do you plan on running any javascript (jest / ...) tests?

StefSchenkelaars: I now 'just' committed your sugestion. But I'm not sure how to write a proper test for this. The ruby tests don't seem to actually compile some code which could be checked. You have any ideas about this?

StefSchenkelaars: @ElMassimo Sorry to be a bit pushy on this one but it currently breaks our build pipeline. The only alternative is to downgrade a lot (including vite to 2.6) so it would be nice if this could be released :)

PR #178, opened 2022-01-15
Remove user-specific entries added to .gitignore

mdesantis:
It removes the `.DS_Store` entry inserted by vite_ruby into `.gitignore`, as it is user-specific and should be added to user's global gitignore.

### DIFF
--- a/vite_ruby/lib/vite_ruby/cli/install.rb
+++ b/vite_ruby/lib/vite_ruby/cli/install.rb
@@ -95,6 +95,8 @@ private
       /public/vite-dev
       /public/vite-test
       node_modules
+      # Vite uses dotenv and suggests to ignore local-only env files. See
+      # https://vitejs.dev/guide/env-and-mode.html#env-files
       *.local
     GITIGNORE
   end
--- a/vite_ruby/lib/vite_ruby/cli/install.rb
+++ b/vite_ruby/lib/vite_ruby/cli/install.rb
@@ -96,7 +96,6 @@ private
       /public/vite-test
       node_modules
       *.local
-      .DS_Store
     GITIGNORE
   end
 


ElMassimo:
The idea is to provide reasonable defaults.

- `node_modules`: Vite Ruby will install npm packages during the installation process
- `.local`: Vite will automatically read environment variables from `.env.local` and similar `.env` files

Makes sense to remove `.DS_Store`, it's already included in most starter `.gitignore` files. Would you update the PR?

mdesantis:
Hi Maximo <3

I disagree about `node_modules`: I can't see why you should set up vite_ruby before setting up any JavaScript package manager. Are you sure about that?

Is the `*.local` behaviour documented somewhere? it surprises me. I wonder why vite_ruby takes responsibility for handling env vars, why don't use something like dotenv instead? :thinking: also, is it documented anywhere?

I'll update the PR if you're still sure about it

ElMassimo:
> why you should set up vite_ruby before setting up any JavaScript package manager?

Rails 7 and other Ruby frameworks do not depend on node, so it's possible that the Vite Ruby installation is the first time you install npm packages in a project.

Vite Ruby supports `npm`, `yarn`, and `pnpm`, all of which install packages in `node_modules` by default. It's bad practice to add that to source control, that's what lockfiles are for.

> I wonder why vite_ruby takes responsibility for handling env vars, why don't use something like dotenv instead? 🤔 

It doesn't. [Vite does, it's well documented, and is powered by `dotenv`.](https://vitejs.dev/guide/env-and-mode.html#env-files)

mdesantis:
Thank you for addressing my concerns, I see your thinking now.

What do think about putting a comment referencing https://vitejs.dev/guide/env-and-mode.html#env-files above the `*.local` line? I would have never realized that without your explanation

ElMassimo:
I like the suggestion, adding a commented link above the `*.local` line 👍

mdesantis:
I updated the PR; feel free to update my changes as you like, if needed!

vfonic:
Awesome! 🤘  Thanks @mdesantis for this PR! I came here to search why would `vite_rails` add `.DS_Store` to my `.gitignore`.

Thanks for removing this! I'll upgrade the gem version in my `Gemfile` now.

Issue #177, opened 2022-01-14
Keep assets from previous builds by default


zealot128:
Reasoning: during "traditional" capistrano-like deployments the 'public/vite' folder is kept between deployments. It is expected to serve old assets for a "little time" between deployments, or for long running clients, that still request old assets.

The docs write about a separate clobber + clean task, so my expectation was, that Vite-Ruby does not delete assets on normal build. But I found out when trying to deploy a new app with Vite, vite:build etc. will delete previous asset versions every time from the 'public/vite/assets', because the option is set by the plugin:

https://github.com/ElMassimo/vite_ruby/blob/91d2ea44873005a53e97a15e76389feccf99106f/vite-plugin-ruby/src/index.ts#L42

This will lead to missing 404 assets during deployment, and is unexpected, as the "normal" rails asset pipeline + webpacker only add new additional assets and a availability of clean/clobber task implies a non-deleting build IMO.

So my suggestion: 

1. Set emptyOutDir to false when in production (like the line below
2. Or under Deployment in docs tell about this option

<details>
<summary>Workaround:</summary>

```javascript
export default defineConfig({ 
  ...
  build: {
    emptyOutDir: false
  }
```

</details>

<details>
<summary>Versions</summary>

```
    "vite": "^2.7.12",
    "vite-plugin-ruby": "^3.0.7",
```

</details>

ElMassimo:
Hi Stefan!

It makes sense for Vite Ruby to depart from the Vite default, and use `emptyOutDir: false` in production if the user does not explicitly provide a value.

```js
emptyOutDir: userConfig.emptyOutDir ?? !isProd,
```

This should be [documented in the Config section](https://vite-ruby.netlify.app/config/#source-maps-%F0%9F%97%BA).

Pull requests are welcome!

ElMassimo:
Closed by #180, thanks!

PR #180, opened 2022-01-18
feat: allow setting emptyOutDir config and set it to false in production mode (#177)

zealot128:
See #177 - assets been deleted between deployments.

Not sure, if there any other places to edit. I guess the ENV variables are populated automatically and transformed to camelcase options?


ElMassimo:
Hi Stefan, thanks for your contribution!

In order to add support for an environment variable, it would be necessary to add it to `default.vite.json`. However, I see environment variables as an useful device for settings that affect both the Ruby portion of the code _and_ `vite-plugin-ruby`.

For this particular feature, users that want to clear the directory can use `vite.config.ts` instead.

PR #199, opened 2022-04-01
fix: disable emptyOutDir in non-development modes

antoinematyja:
### Description 📖

This pull request solves the issue discussed in https://github.com/ElMassimo/vite_ruby/discussions/186
Staging mode was inconsistent with production mode

### Background 📜

This was happening because the code was checking for `mode === 'production'`, but we need the same settings in all non-dev environments (not sure about CI but I guess?).

### The Fix 🔨

Replaced previous `isProduction` with `!isDevelopment`, also added a test but my test game is pretty bad, feel free to rework it

### DIFF
--- a/vite-plugin-ruby/src/index.ts
+++ b/vite-plugin-ruby/src/index.ts
@@ -37,11 +37,11 @@ function config (userConfig: UserConfig, env: ConfigEnv): UserConfig {
   const fs = { allow: [projectRoot], strict: true }
   const server = { host, https, port, strictPort: true, fs }
 
-  const isDevelopment = config.mode === 'development'
+  const isLocal = config.mode === 'development' || config.mode === 'test'
 
   const build = {
-    emptyOutDir: userConfig.build?.emptyOutDir ?? isDevelopment,
-    sourcemap: !isDevelopment,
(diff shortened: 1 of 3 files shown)


vite-plugin-ruby/src/index.ts
```diff
@@ -37,11 +37,11 @@ function config (userConfig: UserConfig, env: ConfigEnv): UserConfig {
   const fs = { allow: [projectRoot], strict: true }
   const server = { host, https, port, strictPort: true, fs }
 
-  const isProduction = config.mode === 'production'
+  const isDevelopment = config.mode === 'development'
```
ElMassimo, line 40: For the purpose of removing old assets in the local environment:

```suggestion
  const isDevelopment = config.mode === 'development' || config.mode === 'test'
```

ElMassimo:
Hi Antoine! @antoinematyja 

Looks great, thanks for your contribution!

ElMassimo:
Released in [`vite-plugin-ruby@3.0.9`](https://github.com/ElMassimo/vite_ruby/blob/main/vite-plugin-ruby/CHANGELOG.md#309-2022-04-01), thanks again!

PR #217, opened 2022-06-01
Catch npm command not found, for folks who install deps with yarn

woodhull:
### Description 📖

This pull request restores the previous behavior where the lack of npm and npx does not cause precompilation to raise.

### Background 📜

This behavior changed in https://github.com/ElMassimo/vite_ruby/commit/967fbf52aac5e52e1a059bffda79c7472874775f.

Previously, it was not necessary to have `npm` installed.

### The Fix 🔨

By changing the code to fallback on error, environments without `npm` will not get an error. 

Long term a better fix might be making pre-installation of dependencies optional for folks who manage their js dependencies using a tool other than npm.

### DIFF
--- a/vite_ruby/lib/tasks/vite.rake
+++ b/vite_ruby/lib/tasks/vite.rake
@@ -42,13 +42,8 @@ namespace :vite do
 
   desc 'Ensure build dependencies like Vite are installed before bundling'
   task :install_dependencies do
-    cmd = begin
-            `npm --version`.to_i < 7 ? 'npx ci --yes' : 'npx --yes ci'
-          # rescue if npm command is not installed to fallback
-          rescue Errno::ENOENT
-            'npx --yes ci'
-          end
-
+    legacy_flag = `npm --version`.to_i < 7 rescue false
+    cmd = legacy_flag ? 'npx ci --yes' : 'npx --yes ci'
     system({ 'NODE_ENV' => 'development' }, cmd)
   end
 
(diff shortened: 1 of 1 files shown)


ElMassimo:
Thanks Nathan!

ElMassimo:
Released in [vite_ruby@3.1.6](https://github.com/ElMassimo/vite_ruby/blob/main/vite_ruby/CHANGELOG.md#316--2022-06-02), thanks!

brendon:
If `yarn` is present in `bin/` in a Rails app then Rails will run `yarn install` already. In my case I get the bundler running twice. I'd prefer to only have the Rails yarn install run. I'm guessing the only way to do this at the moment is to clear this rake task?

ElMassimo:
@brendon Due to all possible combinations in Rails (`sprockets`, `webpacker`, `shakapacker`, `jsbundling-rails`, etc), it's not practical to detect and skip installation.

If you [have configured](https://vite-ruby.netlify.app/guide/deployment.html#using-heroku) your deployment to ensure development dependencies are installed, clearing the `install_dependencies` task is a valid approach.

brendon:
Thanks @ElMassimo, I understand the reluctance to do any kind of detection. Perhaps this could become a configuration option in the future so that it's more visible from the documentation?

ElMassimo:
I have [added](https://github.com/ElMassimo/vite_ruby/commit/c107c28678238c2d9be3ab74a73aa392bbd336ca) more information about dev dependencies and pruning.

I'd rather avoid additional configuration options for this. The current behavior allows to use Vite Ruby in combination with any other library (`propshaft`, `webpacker`, etc), while minimizing the chance of misconfiguration (such as missing development dependencies).

Advanced users can choose to [skip the extension of `assets:precompile`](https://vite-ruby.netlify.app/guide/deployment.html#disabling-extension-of-the-assets-precompile-task) and use `bin/rake vite:build` directly.

PR #238, opened 2022-08-09
fix: use javascript_include_tag in vite_client_tag

CyberAP:
### Description 📖

In order to inherit all the features of the `javascript_include_tag` it should be used within `vite_client_tag`. Otherwise you won't get CSP support and all other built-in features.

### Background 📜

You won't be able to work with `vite_client_tag` in a CSP environment without this change.

### The Fix 🔨

Use `javascript_include_tag` instead of `tag.script`


ElMassimo:
Looks good, thanks!

PR #241, opened 2022-08-22
fix: prevent yarn 2+ error in `assets:precompile`

pftg:
### Description 📖

For Yarn 2+, `YARN_PRODUCTION=false` causes `Usage Error: Unrecognized or legacy configuration settings found: production - run "yarn config -v" to see the list of settings supported in Yarn (in <environment>)`

### Background 📜

Because there is no such global configuration for Yarn 2+, at least I have not found any similar on https://yarnpkg.com/configuration/yarnrc, we got the issue

### The Fix 🔨

Do not add env to prevent warning message if yarn version is more than `1`

### Screenshots 📷

image

### DIFF
--- a/vite_ruby/lib/tasks/vite.rake
+++ b/vite_ruby/lib/tasks/vite.rake
@@ -86,5 +86,8 @@ if ARGV.include?('assets:precompile')
   else
     ENV['NPM_CONFIG_INCLUDE'] = 'dev'
   end
-  ENV['YARN_PRODUCTION'] = 'false'
+
+  unless ViteRuby.commands.next_yarn?
+    ENV['YARN_PRODUCTION'] = 'false'
+  end
 end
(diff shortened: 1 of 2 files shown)


pftg:
@ElMassimo could you check?

ElMassimo:
It's been 3 days Paul, don't be pushy 😄 

I'll check when I have time to review it.

ElMassimo:
[Yarn's migration guide](https://yarnpkg.com/getting-started/migration/#renamed) mentions that selecting which dependencies are installed (dev or prod) requires the `workspace-tools` plugin.

Since a user might not have `workspace` enabled in the project, I think that the change in this pull request makes sense.

This also means it's now the user's responsibility to ensure any build dependencies are available when `vite build` is run.

---

Thanks for your contribution Paul! 😃

Issue #258, opened 2022-09-16
docs: document bug in safari that caches css and prevents hmr


powersurge360:
[troubleshooting section]: https://vite-ruby.netlify.app/guide/troubleshooting.html

- [X] I have tried upgrading by running `bundle update vite_ruby`.
- [X] I have read the __[troubleshooting section]__ before opening an issue.

### Description 📖

In Safari on macOS (at least) when using `vite_stylesheet_tag` it is difficult to get new copies of a stylesheet as it changes. I believe this is because Safari isn't respecting the `Cache-Control: no-cache` when an `ETag` header is set. I am unsure if this is a bug in rails, vite-ruby or vite-rails. I was unable to work out where the `ETag` is set. I also tried in Chrome and there didn't seem to be an issue.

### Reproduction 🐞

_Please provide a link to a repo that can reproduce the problem you ran into._

<details>
<summary>Vite Ruby Info</summary>

_Run `bin/rake vite:info` and provide the output:_

```

```
</details>

### Logs 📜

_If not providing a reproduction:_

<details>
<summary>Output</summary>

_Run `DEBUG=vite-plugin-ruby:* bin/vite dev` or `DEBUG=vite-plugin-ruby:* bin/vite build` and provide the output:_

```
bin/vite present?: true
vite_ruby: 3.2.3
vite_rails: 3.0.12
rails: 7.0.3.1
node: v16.17.0
npm: 8.15.0
yarn: 1.22.19
pnpm: 7.5.2
ruby: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21]

installed packages:
pimiibo@ /Users/powersurge360/projects/pimiibo
├─┬ vite-plugin-ruby@3.1.2
│ └── vite@3.1.0 deduped
└── vite@3.1.0
```
</details>

### Screenshots 📷

_Provide console or browser screenshots of the problem_.

On first load
<img width="1511" alt="image" src="https://user-images.githubusercontent.com/1203658/190829149-798ea0b6-4b59-4d49-9f6d-f5c8e687a9a8.png">

On second load

<img width="1511" alt="image" src="https://user-images.githubusercontent.com/1203658/190829201-83e6b2db-8446-4164-bca5-b0d7a07f8e0b.png">

On manually clearing caches and reloading again
<img width="1512" alt="image" src="https://user-images.githubusercontent.com/1203658/190829244-8140ae1c-a1a5-400b-95e8-87b79805708f.png">

[...]

PR #261, opened 2022-09-20
docs: document bug in safari that caches css and prevents hmr

powersurge360:
Discussed in [an issue](https://github.com/ElMassimo/vite_ruby/issues/258) opened the other day.

### Description 📖

Safari does not respect the combination of `ETag` and `Cache-Control: no-cache` headers without the use of a cache-busting query string. The `vite_stylesheet_tag` helper does not have this cache-busting query string but if you import the CSS directly into the JS, when it gets split out it works.

### Background 📜

Bug in Safari

### The Fix 🔨

We are documenting the behavior to help folks who may get stumped by it. In the issue, another option was linked in the comments as a workaround but that option didn't appear in my version of Safari and I wasn't able to work out how to disable cache. The only other option I've found so far is to manually empty the cache between stylesheet changes which is not a great dev experience.

### Screenshots 📷

Not relevant to this change, just light documentation adjustments.

### DIFF
--- a/docs/src/guide/troubleshooting.md
+++ b/docs/src/guide/troubleshooting.md
@@ -157,6 +157,10 @@ This is probably the case if you are seeing errors such as `#<Errno::EMFILE: Too
 
 Follow [this article][ulimit] for information on how to increase the limit of file descriptors in your OS.
 
+### CSS does not seem to reflect changes in dev
+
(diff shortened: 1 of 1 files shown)


powersurge360:
I stuck it in network because it made the most sense to me of the existing headers. Let me know if you'd like that put somewhere else or want more background in the description of the issue in the markdown file.

ElMassimo:
Thanks Kevin! I'll make some light edits after merging, but this looks great!

PR #278, opened 2022-10-16
fix: yarn berry pnp mode support

heka1024:
---

Thank you for this wonderful gem! I really love it ❤️ 

### Description 📖

This pull request supports yarn berry's pnp mode.

### Background 📜

When I used vite-ruby with yarn berry pnp mode, there was an error about `bin/vite dev`. You can reproduce it in this repository. https://github.com/heka1024/yarn_berry_vite. The problem was `vite_ruby` looked up `node_modules/.bin/vite` but there wasn't `node_modules`.

### The Fix 🔨

By changing `vite_ruby/lib/vite_ruby/runner.rb` to use `yarn` instead of `node_modules/.bin/vite`, we can now use the zero install mode of yarn berry. 😃 

I think 
```ruby
if ViteRuby.commands.yarn_berry?
  cmd.push('yarn', 'vite')
else
  cmd.push(vite_executable)
end
```
this is ugly, but I can't imagine a better one. I'd appreciate it if you could suggest a better way.

### DIFF
--- a/vite_ruby/lib/vite_ruby/runner.rb
+++ b/vite_ruby/lib/vite_ruby/runner.rb
@@ -31,7 +31,11 @@ private
       args = args.clone
       cmd.push('node', '--inspect-brk') if args.delete('--inspect')
       cmd.push('node', '--trace-deprecation') if args.delete('--trace_deprecation')
-      cmd.push(vite_executable)
+      if ViteRuby.commands.yarn_berry?
+        cmd.push('yarn', 'vite')
+      else
+        cmd.push(vite_executable)
+      end
       cmd.push(*args)
       cmd.push('--mode', config.mode) unless args.include?('--mode') || args.include?('-m')
     end
--- a/vite_ruby/lib/vite_ruby/commands.rb
+++ b/vite_ruby/lib/vite_ruby/commands.rb
@@ -71,9 +71,19 @@ class ViteRuby::Commands
     `npm --version`.to_i < 7 rescue false
   end
 
+  # Internal: Checks if the package manager is yarn.
+  def yarn?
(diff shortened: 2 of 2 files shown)


ElMassimo:
Hi Koo!

Thanks for the detailed PR!

vite_ruby/lib/vite_ruby/runner.rb
```diff
@@ -31,7 +31,11 @@ def command_for(args)
       args = args.clone
       cmd.push('node', '--inspect-brk') if args.delete('--inspect')
       cmd.push('node', '--trace-deprecation') if args.delete('--trace_deprecation')
-      cmd.push(vite_executable)
+      if ViteRuby.commands.yarn_berry?
+        cmd.push('yarn', 'vite')
+      else
+        cmd.push(vite_executable)
+      end
```
ElMassimo, line 38: Let's move this check to `vite_executable`, modifying it to return an array:

```ruby
cmd.push(*vite_executable)
```

We would still want a user to be able to configure `vite_bin_path` as needed:

```ruby
  # Internal: Resolves to an executable for Vite.
  def vite_executable
    bin_path = config.vite_bin_path
    return [bin_path] if File.exist?(bin_path)
 
    if config.root.join('yarn.lock').exist?
      ['yarn', 'vite']
    else
      ["#{ `npm bin`.chomp }/vite]
    end
  end
```

That way we can more easily extend this for other package managers that don't use a `node_modules` structure.

heka1024: Thank you for comment! 👍

ElMassimo:
Thanks for your contribution!

PR #325, opened 2022-12-22
chore: bump dry-cli version

vladimirtemnikov:
### Description 📖

Hello!
Could we bump `dry-cli` dependency version to `1.0.x`? Looks like no breaking changes from `~0.7`

### DIFF
--- a/vite_ruby/vite_ruby.gemspec
+++ b/vite_ruby/vite_ruby.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = Gem::Requirement.new('>= 2.5')
 
-  s.add_dependency 'dry-cli', '~> 0.7.0'
+  s.add_dependency 'dry-cli', '~> 1.0'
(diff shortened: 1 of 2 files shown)


vite_ruby/vite_ruby.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = Gem::Requirement.new('>= 2.5')
 
-  s.add_dependency 'dry-cli', '~> 0.7.0'
+  s.add_dependency 'dry-cli', '~> 1.0'
```
ElMassimo, line 21: Let's allow `0.7`, as there are no breaking changes.

vladimirtemnikov: Done. Is it ok to use `< 2`? Or we should prefer `< 1.1`?

ElMassimo: `< 2` should be fine now that it's a major release. Thanks!

PR #337, opened 2023-02-07
feat: allow javascript_include_tag options to vite_client_tag

mjhoy:
### Description 📖

This allows passing options such as `nonce`, which some browsers require if CSP is set (Firefox).

### Background 📜

With CSP set in Rails, Firefox would fail to load the vite client tag without a `nonce` set.

### The Fix 🔨

`vite_client_tag` now accepts an `options` hash that is passed to `javascript_include_tag`, much like `vite_javascript_tag`.

Co-authored-by: Wil Hall <wil@wilhall.com>


ElMassimo:
Thanks Mikey and Wil!

Issue #249, opened 2022-09-05
Add nonce to `vite_react_refresh_tag` Rails tag helper 


renchap:
### Description 📖

When using a Content Security Policy that does not allow `unsafe-inline` scripts, `vite_react_refresh_tag` does not work as inline JS is not allowed. The solution is to enable CSP nonces (Rails has a way to do it), but then the nonce is not added to the generated `<script>` tag and the tag is ignored.

Note that this is quite common policy using nonces or script hashes, [needed for the legacy plugin](https://github.com/vitejs/vite/tree/main/packages/plugin-legacy#content-security-policy) for example

### Solution 💡

I see 2 solutions here, and I can try a PR to implement the one you prefer:
- Probably best solution: update `react_refresh_preamble` to only return the JS content, and wrap it into the framework specific tag for each framework, allowing to pass framework options as needed (for example `vite_react_refresh_tag(nonce: true)`)
- update `vite_react_refresh_tag` and `react_refresh_preamble` to accept a `nonce` argument, and add it to the generated HTML

What do you think?

ElMassimo:
Hi Renaud, thanks for reporting.

Ideally, we add a new `react_preamble_code` method in `ViteRuby::Manifest`  to preserve backwards compatibility, keeping `react_refresh_preamble`, and then extend `vite_rails` to use `javascript_tag` with `nonce: true, type: :module`.

PRs are welcome!

mikecx:
Hey, just wondering if there's been any progress on this. We're looking to get CSP enabled and this is obviously blocking the more advanced CSP cases.

ElMassimo:
This only affects development, where you can make an exception if needed.

> PRs are welcome!

PR #372, opened 2023-06-13
feat: Use javascript_tag helper for vite_react_refresh_tag

omarowns:
feat: Use javascript_tag helper for vite_react_refresh_tag allowing h…tml_options

### Description 📖

This pull request updated the `vite_react_refresh_tag` method to use the Rails JavaScript helper method [`javascript_tag`](https://api.rubyonrails.org/classes/ActionView/Helpers/JavaScriptHelper.html#method-i-javascript_tag) in order to support HTML options.

### Background 📜

This was happening because of #249. When adding Content Security Policy on a Rails project with something other than `unsafe-inline` the inline JS for the react refresh script was not being allowed.

### The Fix 🔨

By changing the `vite_react_refresh_tag` to use Rails' `javascript_tag` we can now add HTML options such as `nonce: true`.

The code adds the method `react_preamble_code` which has the script for the React Refresh plugin.

The `react_refresh_preamble` is left unchanged with the exception of using the `react_preamble_code` method internally inside its `<script>...</script>` tags.

### DIFF
--- a/test/helper_test.rb
+++ b/test/helper_test.rb
@@ -59,6 +59,10 @@ end
 class HelperTest < HelperTestCase
   tests ViteRails::TagHelpers
 
+  def content_security_policy_nonce
+    'iyhD0Yc0W+c='
+  end
+
   def test_vite_client_tag
     assert_nil vite_client_tag
     with_dev_server_running {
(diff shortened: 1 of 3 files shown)


ElMassimo:
Looks great, thanks for your contribution Omar!

PR #374, opened 2023-06-19
Allow skipping dev dependencies on install

eriknygren:
### Description 📖

Add an env flag `VITE_RUBY_SKIP_INSTALLING_DEV_DEPENDENCIES`, which lets users skip installing dev dependencies in the `install_dependencies` rake task

Not sure what's the best way to test this, so opening as a draft to ask for advice, and see if anyone else thinks this is useful. For my project `devDependencies` only includes non essential things like storybook and linters, hence why I'd like the option to skip installing them

### Background 📜

For projects where dev dependencies only include linters etc, this is a nice way of speeding up the build. Similar to how on legacy yarn you could run `yarn install --production`

### The Fix 🔨

Alter the task to optionally add `--omit=dev` to the `npx ci` build command.

### DIFF
--- a/vite_ruby/lib/tasks/vite.rake
+++ b/vite_ruby/lib/tasks/vite.rake
@@ -42,11 +42,11 @@ namespace :vite do
 
   desc 'Ensure build dependencies like Vite are installed before bundling'
   task :install_dependencies do
+    install_env_args = ENV['VITE_RUBY_SKIP_INSTALL_DEV_DEPENDENCIES'] == 'true' ? {} : { 'NODE_ENV' => 'development' }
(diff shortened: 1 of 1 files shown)


ElMassimo:
Hi Erik!

Sounds reasonable, though it should be enough to set `NODE_ENV` accordingly.

If the flag is set, we would omit `NODE_ENV`. Otherwise, we pass `NODE_ENV: 'development'` as before.

eriknygren:
Hi @ElMassimo, thanks for the feedback, I updated the PR, is that what you meant?

ElMassimo (approved):
Looks good, thanks!

eriknygren:
Nice one @ElMassimo, I just squashed that into on commit, thanks for the extremely quick turnaround, much appreciated ⭐

PR #377, opened 2023-06-26
fix: Check only once per second when dev server is not running

mattbrictson:
### Description 📖

This pull request speeds up performance considerably in local test environments by reducing the frequency of the dev server check when the dev server is not running.

### Background 📜

The comment for the `dev_server_running?` method says:

> Checks only once every second

However, in reality that is only true with the dev server is running.

When the dev server is _not_ running, no timestamp is recorded. That means that the dev server check is performed with every manifest lookup.

### The Fix 🔨

Store a timestamp in `@running_checked_at` regardless of whether the dev server check succeeds or fails. This ensures that the dev server check is done at most once per second, even if the dev server is not running.

### Screenshots 📷

In a Rails app with a large number of images and stylesheets (i.e. many manifest lookups per page), this patch resulted in an 18x performance improvement when running system specs.

Before:

```
Finished in 4 minutes 15.7 seconds (files took 2.93 seconds to load)
22 examples, 0 failures
```

After:

```
Finished in 14.2 seconds (files took 2.84 seconds to load)
22 examples, 0 failures
```

### DIFF
--- a/test/dev_server_test.rb
+++ b/test/dev_server_test.rb
@@ -8,13 +8,18 @@ class DevServerTest < ViteRuby::Test
 
     refresh_config(mode: 'development')
     refute ViteRuby.instance.dev_server_running?
+
+    running_checked_at = ViteRuby.instance.instance_variable_get('@running_checked_at')
+    assert_in_delta(Time.now.to_f, running_checked_at.to_f, 0.01)
   end
 
   def test_running
     refresh_config(mode: 'development')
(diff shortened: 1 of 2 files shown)


ElMassimo:
> In a Rails app with a large number of images and stylesheets

Makes sense! Thanks!

Disabling `auto_build` in the CI would bail the check early before attempting a socket connection, so hadn't noticed it.

This will make the development server check a bit less responsive, but it should be fine in practice.

Issue #324, opened 2022-12-22
Make JS package manager configurable (vs. detected)


tkbrigham:
**Is your feature request related to a problem? Please describe.**
When I run `bundle exec vite install` in my pnpm subproject (which is a Rails app), I get the following:
```
Installing js dependencies
npm ERR! code ENOTSUP
npm ERR! notsup Unsupported engine for rollup@3.7.5: wanted: {"node":">=14.18.0","npm":">=8.0.0"} (current: {"node":"14.20.0","npm":"6.14.11"})
npm ERR! notsup Not compatible with your version of node/npm: rollup@3.7.5
npm ERR! notsup Not compatible with your version of node/npm: rollup@3.7.5
npm ERR! notsup Required: {"node":">=14.18.0","npm":">=8.0.0"}
npm ERR! notsup Actual:   **{"npm":"6.14.11","node":"14.20.0"}
```

It is looking at npm versions, even though my pnpm is updated. I believe this is because the subproject (which I believe is the config.root) does not detect a pnpm-lock.yaml file ([source](https://github.com/ElMassimo/vite_ruby/blob/eb8f678248a02b693fffe5a49309984fed92a051/vite_ruby/lib/vite_ruby/cli/install.rb#L121)). pnpm-lock.yaml does exist one directory above my subproject, in my repo (and pnpm workspace) root.

**Describe the solution you'd like**
It would be easier to allow a user to pass in their preferred package manager as a CLI argument (or ENV var) when running `bundle exec vite install` versus relying on detection. The detection is great as a default, but should be configurable/overridable by the user.

**Describe alternatives you've considered**
None, but happy to do so if it would help. When I hardcode my preferred `pnpm install` into the function below, `bundle exec vite install` works as expected. I'm happy to submit the PR if this suggestion seems reasonable!

**Additional context**
Context linked above, but the `def npm_install` method is probably the main place I would look to make changes, along with some command line argument parsing. Relevant function: https://github.com/ElMassimo/vite_ruby/blob/eb8f678248a02b693fffe5a49309984fed92a051/vite_ruby/lib/vite_ruby/cli/install.rb#L118-L124

ElMassimo:
Hi Thomas!

The detection logic could be improved by using [`pkg-up`](https://github.com/sindresorhus/pkg-up) to detect the project root before checking the lockfiles.

Passing the preferred package manager with a `--with` flag would also be a nice improvement:

```
bundle exec vite install --with pnpm
```

[...]

PR #392, opened 2023-08-05
Adds Bun support

pftg:
### Description 📖

Adds support of Bun

### Background 📜

Lately, Bun has adopted Vite, so we can use it for compilation assets.


vitalyliber:
It will be great to see the Bun support!

pftg:
@vitalyliber left to fix tests ;( but you can use my pr to confirm at least this "fast" workaround works for you too

pftg:
@ElMassimo please review this PR

### Notes:

This is a very naive approach to start using it. I already use it for my projects.
Later, I saw some ideas to encapsulate all those package managers: remove duplication to find the current package manager.

pftg:
I found some strange issue: `(!) "base" option should start with a slash.`

vladimirtemnikov:
Hello, @pftg 

Currently, bun does not support brotli: https://github.com/oven-sh/bun/issues/267

ViteRuby could work with bun, but we have [vite-plugin-rails](https://github.com/ElMassimo/vite_ruby/tree/main/vite-plugin-rails ), which depends on brotli compression currently, may be we should wait some time before brotli will be implemented or conditionally enable brotli in plugin?

@ElMassimo any thoughts on this?

pftg:
@vladimirtemnikov Yeah, you are right.

I had to disable brotli by the way, to use this version. So if this is a hard requirement to support the compress plugin, then there is a workaround.

So, for now, I use `compress: { brotli: false }` to use a bun. I'm ok with that ;)

I think better to start using without brotli to find more issues

pftg:
@vladimirtemnikov @ElMassimo updated PR, what do you think to add notice in the Readme that it's not compatible with the vite rails plugin and list of supported bun versions?

pftg:
Hmm, for development mode, everything was fine, but as soon as I tried to run on Heroku with production mode, the compilation got stuck. I found that when removing RubyPlugin, it's not stuck. I need help to set env to see the real problem. I'll try to look into it later.

Will be great to find a guide on how to set up development env for those plugins with bun and with integration testing in the real app.

dominikklein:
Any ideas to bring this forward? Also rails 7.1. has now bun support. Would be nice to when it can also be used in the vite-ruby/rails context.

pftg:
Yeah need to check latest tests, maybe now it's working. For now I'm using this approach to use bun https://jtway.co/integrating-bun-with-vite-ruby-for-lightning-fast-frontend-builds-bbd220fefdc0

pftg:
@ElMassimo, with the latest bun, it looks like it works. I'm working on the cleanup in separate PR

vite_ruby/lib/vite_ruby/package_manager.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module ViteRuby::PackageManager
+  def self.resolve(root:)
+    package_manager_name = ENV.fetch('VITE_RUBY_PACKAGE_MANAGER', detect_package_manager(root))
+    package_manager_class_for(package_manager_name).new(root:)
```
schonert, line 6: @pftg Missing the actual reference to the `root` variable. This is what's breaking the builds.
Should be:
```rb
package_manager_class_for(package_manager_name).new(root: root)
```

schonert: Okay - managed to pull down the project and fix the error which then lead to more build errors. Not sure where to go from here

pftg:
https://github.com/ElMassimo/vite_ruby/actions/runs/8186414098/job/22384757347?pr=392 - it's a new issue that appeared lately, I will check, but I believe I need to rebase

vite_ruby/lib/vite_ruby/package_manager.rb
pftg: > @pftg Missing the actual reference to the `root` variable. This is what's breaking the builds. Should be:
> 
> ```ruby
> package_manager_class_for(package_manager_name).new(root: root)
> ```

Oh, I forgot about other ruby versions. You are right

pftg: Hey @schonert, how can I trigger running all checks?

schonert:
@pftg all green - great job!

adamburmister:
Fantastic, just what I need, but I see this has sat around in a green state for a month+. 
Is there a plan to merge and publish this?

ElMassimo:
Hi @pftg and other folks!

I've added support for a [`packageManager` config option](https://vite-ruby.netlify.app/config/index.html#packagemanager):

- https://github.com/ElMassimo/vite_ruby/issues/324

As part of that feature, it felt it was reasonable to add experimental support for `bun`, as it only required [two](https://github.com/ElMassimo/vite_ruby/blob/main/vite_ruby/lib/vite_ruby/runner.rb#L46) [lines](https://github.com/ElMassimo/vite_ruby/blob/main/vite_ruby/lib/vite_ruby/config.rb#L114) of code.

Give it a try in `vite_ruby-3.7.0`.

The implementation in this pull request is reasonable, but I'd like to minimize the surface area of the library (users typically end up using internals even if they are marked as such).

pftg:
@ElMassimo, could you clarify whether we should close this PR or if you need me to rework some parts?

pftg:
Oh, it just adds settings to avoid auto-detection. In this pr, you find encapsulations of package managers at all, and I also added package manager settings too ;) https://github.com/ElMassimo/vite_ruby/pull/392/files#diff-756bec63bbf1c84f70817c5a763e016026db97b00b4b5e99a074dfc2ddb9521cR5

Shaglock:
Hi @ElMassimo, are there any plans to merge this or make vite_ruby not depend on NodeJS being installed in the system?

pftg:
@ElMassimo I forgot to mentione that there are missed more lines, you can check my PR what other changes need to add to make bun works. But I'd suggest to cleanup once and remove duplications with managing js packages, it was a little hard to find all those parts and cleanup it.

danielnc:
@ElMassimo it would be nice to get this merged and how to enable bun with vite_rails we are using bun to  compile assets in prod and works fine but dev still relies on node

ElMassimo:
[my reply]: https://github.com/ElMassimo/vite_ruby/pull/392#issuecomment-2233463715

To clarify what I meant with:

> [I'd like to minimize the surface area of the library][my reply]

I'm not interested in merging this refactoring, I'd prefer to keep the library lighter (less classes, smaller footprint).

Feel free to open an issue/PR if [setting `"packageManager": "bun"`](https://vite-ruby.netlify.app/config/index.html#packagemanager) is not working as intended, providing error traces or relevant system information.

---

Given the misunderstanding after [my reply], I've updated the [CONTRIBUTING](https://github.com/ElMassimo/vite_ruby/blob/main/CONTRIBUTING.md#new-features) docs to minimize the chance of similar situations in the future.

A few [notes from it](https://github.com/ElMassimo/vite_ruby/blob/main/CONTRIBUTING.md#new-features):
- If you’d like to discuss a new feature, [create a new Discussion](https://github.com/ElMassimo/vite_ruby/discussions/new?category=ideas).
- Keep changes to a minimum. If you have multiple features or fixes, submit multiple pull requests.
- Please open a discussion to get feedback on your idea before spending too much time on it.

PR #413, opened 2023-10-23
Respect host and protocol options in vite_preload_tag

rywall:
### Description 📖

When specifying the host and protocol options for a `vite_javascript_tag`, we want the preload tags to also respect those options.

### Background 📜

This was causing an issue where the preload tags would not use the CDN specified for the `vite_javascript_tag`.

### DIFF
--- a/test/helper_test.rb
+++ b/test/helper_test.rb
@@ -105,6 +105,14 @@ class HelperTest < HelperTestCase
     }
   end
 
+  def test_vite_preload_tag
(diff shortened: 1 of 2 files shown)


sleistner:
Hi @ElMassimo,
Just wanted to kindly follow up on this PR. It’s been a while (about a year now), and we’d appreciate your review when you have a moment. Please let me know if any changes or further discussions are needed to move it forward.
Thank you for your time!

sleistner:
Awesome! 👍🏻

PR #421, opened 2023-11-16
feat: add support for vite 5, which changed default manifest path

ElMassimo:
### Description 📖

This pull request adds support for Vite 5 in `vite_ruby`.

It also modifies `vite-plugin-ruby` to output its manifest to `.vite/manifest-assets.json`, in alignment with Vite.

### Background 📜

In Vite 5, the manifest [has a new location, under `.vite/manifest.json`](https://vitejs.dev/guide/migration#manifest-files-are-now-generated-in-vite-directory-by-default).

To avoid breaking existing builds which might be cached, we will still search for manifests in the old locations, during a transitional period.


franzliedke:
Did this only *add* support for Vite 5? It looks like it also removed support for older versions, correct?

PR #440, opened 2024-03-06
build: adds ruby 3.3 and rails 7.1 to CI

pftg:
### Description 📖

There are new Ruby and Rails versions which will be great to check, because of some problems with them

### The Fix 🔨

Adds ruby 3.3 and rails 7.1


larouxn:
Would be great to see this merged. 👏

ElMassimo:
Thanks Paul!

PR #442, opened 2024-03-18
Add `vite.config.mjs` to watched paths

janko:
### Description 📖

This ensures building isn't skipped when `.mjs` extension is used for the Vite config file.

### Background 📜

Vite 5 deprecated CJS, so unless `"type": "module"` is specified in `package.json` (which requires using `.cjs` extension things like PostCSS config), the only way to avoid the deprecation warning is to rename the config file to `vite.config.mjs`.

https://vitejs.dev/guide/migration.html#deprecate-cjs-node-api

### The Fix 🔨

I added `vite.config.mjs` to default watched paths.

### DIFF
--- a/vite_ruby/lib/vite_ruby/config.rb
+++ b/vite_ruby/lib/vite_ruby/config.rb
@@ -195,6 +195,7 @@ private
     postcss.config.js
     tailwind.config.js
     vite.config.js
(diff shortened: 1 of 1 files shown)


ElMassimo:
Thanks Janko!

PR #451, opened 2024-04-02
feat: allow skipping dependency install in assets:precompile

josephnle:
### Description 📖

We want to use the `assets:precompile` extension but we don't want it to install dependencies. This allows setting `VITE_RUBY_SKIP_ASSETS_PRECOMPILE_INSTALL=true` to skip installation while keeping the rest of the task extensions.

### Background 📜

This was happening because dependencies are always installed (`yarn install` / `npx ci`) as part of the `assets:precompile` extension.

### The Fix 🔨

Add conditional controlled by `VITE_RUBY_SKIP_ASSETS_PRECOMPILE_INSTALL` to allow for skipping the installation task.


ElMassimo:
Hi Joseph!

This has been requested enough times that I feel it justifies adding, especially given that it's not unusual for `node_modules` to be installed in a previous layer when creating Docker images.

Thanks!

PR #452, opened 2024-04-04
Fix for "uninitialized constant ViteRuby::Manifest::OpenStruct" Error

aki77:
### Description 📖

This pull request specifically targets an issue occurring only when encountering `ViteRuby::MissingEntrypointError`. 
The problem stems from the absence of `require 'ostruct'` in these error scenarios. 
This fix ensures that `require 'ostruct'` is included in the appropriate context, resolving the issue and preventing errors during `ViteRuby::MissingEntrypointError` occurrences.


ElMassimo:
Thanks!

PR #458, opened 2024-05-01
fix: use correct dirname for manifest

josephnle:
### Description 📖

Uses `path.dirname` instead of `path.basename` so we correctly determine the `manifestDir` in the assets-manifest plugin.

### Background 📜

This was happening because the package was using `path.basename`, which gives the file name and not the directory name.

### The Fix 🔨

Change `path.basename` to `path.dirname`

### Screenshots 📷

N/A


ElMassimo:
Thanks Joseph!

PR #468, opened 2024-06-07
Update `actions/checkout` to v4 (node 16->20)

mjankowski:
Gets rid of deprecation noise in some of the CI runs.


ElMassimo:
Thanks!

PR #469, opened 2024-06-07
Use `SECRET_KEY_BASE_DUMMY` feature as placeholder during build

mjankowski:
I think this will fix the very bottom (Rails edge run) here - https://github.com/ElMassimo/vite_ruby/actions/runs/9422889978/job/25960116922

Rails after 7.1 has switched to expect the secret_key_base to come from credentials, so the secret_key_base config setting from the dummy apps doesn't get picked up, hence that error. This env var should trigger a dummy value to be used (see note in bottom of https://guides.rubyonrails.org/asset_pipeline.html#local-precompilation)


mjankowski:
Yes, looks like this passed the previously failing rails edge run - https://github.com/ElMassimo/vite_ruby/actions/runs/9423464051/job/25961896030

ElMassimo:
Excellent, thank you Matt!

PR #472, opened 2024-06-08
Add `vite_picture_tag` coverage for pre-7.1 test paths

mjankowski:
Previously this method was going untested in pre-7.1 rails test executions -- changes to assert what we expect (raise) in that path.


ElMassimo:
Thanks Matt!

PR #475, opened 2024-06-24
Add missing `.ts` extension in `application.js` example

osbre:
### Description 📖

In the latest Vite, the default falls to `.js`, therefore `.ts` needs to be explicitly specified.

### The Fix 🔨

By updating the example to include the `.ts` file extension


ElMassimo:
Hi Ostap!

This is not necessary, as the helper is provided by Vite Ruby explicitly for this purpose, and will [add the `.ts` extension](https://github.com/ElMassimo/vite_ruby/blob/main/vite_ruby/lib/vite_ruby/manifest.rb#L210).

PR #476, opened 2024-07-02
docs: More robust solution to Safari dev cache bug

mattbrictson:
### Description 📖

Safari had an annoying bug where it will ignore changes to CSS files, breaking the developer's hot-reload experience. The Vite Ruby troubleshooting guide [recommends](https://github.com/ElMassimo/vite_ruby/blob/4433471406e7e4957e13acee5151829b36a8eb60/docs/src/guide/troubleshooting.md#L161) moving CSS into the JS entrypoint as a workaround.

However, in Safari 17.5, the problem is now even worse: the bug now affects JS files as well. That means the previously documented workaround is no longer a solution for the entire scope of Safari HMR issues.

### Background 📜

Ultimately, the problem is a known bug in Safari where it aggressively caches JS and CSS files that are marked `rel=preload`.

> If a request is preloaded, Safari will always retrieve it from the cache, regardless of the cache headers for that request.

https://bugs.webkit.org/show_bug.cgi?id=193533

### The Fix 🔨

Update the Vite Ruby troubleshooting guide with a more robust workaround: disable the `rel=preload` Link header in development. This completely solves the issue for me.

FWIW, I'm running:

macOS 14.5 (23F79)
Safari 17.5 (19618.2.12.11.6)

### DIFF
--- a/docs/src/guide/troubleshooting.md
+++ b/docs/src/guide/troubleshooting.md
@@ -158,11 +158,23 @@ This is probably the case if you are seeing errors such as `#<Errno::EMFILE: Too
 
 Follow [this article][ulimit] for information on how to increase the limit of file descriptors in your OS.
 
-### Safari does not reflect CSS changes in development
+### Safari does not reflect CSS and JS changes in development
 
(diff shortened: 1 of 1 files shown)


mattbrictson:
Looks like there is a global way to disable `rel=preload` that is supported all the way back to Rails 6.1. I'll revise my PR to recommend this approach instead.

ElMassimo:
Looks great, thanks Matt!

Issue #404, opened 2023-09-25
`assets:clean` incorrectly removes files created in the same build


KevinBerthier:
### Description 📖

Some assets are missing on production after deployment. It happens totally randomly. A redeployment fixes the issue

### Reproduction 🐞

deploy.rb
```rb
set :keep_releases, 5
set :use_sudo, false

set :log_level, :debug

set :linked_files, %w[config/database.yml config/application.yml config/secrets.yml public/.htaccess]
set :linked_dirs, %w[.bundle log storage public/system public/uploads downloads node_modules tmp/pids tmp/cache tmp/sockets tmp/dumps]

set :bundle_jobs, 2
set :keep_assets, 2
set :assets_prefix, 'vite'
```

vite.config.ts
```ts
import { resolve } from 'path'
import { defineConfig } from 'vite'
import VuePlugin from '@vitejs/plugin-vue'
import { visualizer } from "rollup-plugin-visualizer";
import ViteRails from 'vite-plugin-rails'
import legacy from '@vitejs/plugin-legacy'



export default defineConfig({
  plugins: [
    legacy({ renderLegacyChunks: false }), // reads .browserlistrc
    ViteRails({
      stimulus: false,
      fullReload: {
        additionalPaths: ['app/**/*', 'config/locales/**/*']
      },
    }), VuePlugin(), visualizer()
  ],
  resolve: {
    alias: {
      '@assets': resolve(__dirname, 'app/assets'),
      '@stylesheets': resolve(__dirname, 'app/assets/stylesheets'),
      '@images': resolve(__dirname, 'app/assets/images'),
      '@pdp': resolve(__dirname, 'app/javascript/apps/pdp'),
    },
  },

})
```

<details>
<summary>Vite Ruby Info</summary>

```
bin/vite present?: true
vite_ruby: 3.3.4
vite_rails: 3.0.15
rails: 7.0.7.2
node: v20.5.1
npm: 9.8.0
yarn: 1.22.17
pnpm: 
ruby: ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [arm64-darwin22]

installed packages:
graphiste@0.1.0 /Users/kevinberthier/code/KevinBerthier/graphiste
├─┬ @vitejs/plugin-legacy@3.0.2
│ └── vite@4.4.9 deduped
├─┬ @vitejs/plugin-vue@4.2.3
│ └── vite@4.4.9 deduped
├─┬ vite-plugin-rails@0.1.0
[...]

PR #477, opened 2024-07-04
fix: Retain bundled current files (close #438)

### LINKED ISSUE #438: clean command not detecting versions properly
### Description 📖

`vite_ruby` uses the `mtime` as a reference to detect different versions of the app: https://github.com/ElMassimo/vite_ruby/blob/main/vite_ruby/lib/vite_ruby/commands.rb#L148 That being said, a single Vite build can generate files with more than one `mtime`. For example, `vite-plugin-rails` uses `rollup-plugin-gzip` at the end of the build which will write many files with different `mtime`s.

This causes files to delete unexpectedly when running `bin/vite clean`. It's possible this bug is hard to reproduce with a small project as all written files may end up having the same `mtime`.

### Reproduction 🐞

<details>
<summary>Vite Ruby Info</summary>

- Run `bin/rake vite:build` on a big project with no currently built assets.
- Run `bin/rake vite:clean[1,0]`

Result: files will be deleted while there is only a single version of the built assets.
</details>

benlangfeld:
This was broken in https://github.com/ElMassimo/vite_ruby/commit/8a581c15ff480049bbb14dab1b5a3497308521b5 with no explanation of why it was being removed.

See #404 and #438 .

### DIFF
--- a/test/commands_test.rb
+++ b/test/commands_test.rb
@@ -41,6 +41,8 @@ class CommandsTest < ViteRuby::Test
 
       js_file = config.build_output_dir.join('assets/application.js')
       js_file.write('export {}')
+      css_module = config.build_output_dir.join('assets/styles.css')
+      css_module.write('.foo {}')
 
       # Simulate using vite-plugin-rails & rollup-plugin-gzip to produce
       # source maps, gzip & brotli compressed versions of the file.
@@ -50,6 +52,10 @@ class CommandsTest < ViteRuby::Test
       gzip_file.write('export {}')
       brotli_file = config.build_output_dir.join('assets/application.js.br')
       brotli_file.write('export {}')
+      css_gzip_file = config.build_output_dir.join('assets/styles.css.gz')
+      css_gzip_file.write('.foo {}')
+      css_brotli_file = config.build_output_dir.join('assets/styles.css.br')
+      css_brotli_file.write('.foo {}')
 
       # Should not clean, the manifest does not exist.
       refute clean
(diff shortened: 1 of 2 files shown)


philippevezina:
@benlangfeld, you should remove the "Closes" keyword as it does not actually fix the reported issue: https://github.com/ElMassimo/vite_ruby/issues/438#issuecomment-2210934756

benlangfeld:
@ElMassimo Any chance this could be reviewed? It fixes a critical bug in this gem and deserves to be released urgently.

ElMassimo:
I'd like to merge a fix that fully addresses #438.

The main problem is that—unlike Webpack—Rollup will create files with different `mtime`, so we should cluster them in order to have `clean` with the default parameters (used implicitly through `assets:precompile`) working as expected in most scenarios.

Leveraging the manifest to keep referenced files is nice (your patch fixes the current implementation), but it's not sufficient to address the problems in the current `clean` implementation.

I'm also considering removing `clean` altogether, as it has the potential of breaking existing deployments, and let each application handle server-based deployments as needed. A separate gem that is tailored for `capistrano` or similar deployment tools can easily extend `assets:clean`, so this doesn't necessarily need to live in `vite_ruby`.

benlangfeld:
@ElMassimo Does a more complete fix exist? Assuming it does not, why not merge this in the meantime? Right now the implementation is fundamentally broken and dangerous.

benlangfeld:
Well, looks like this feature was removed in https://github.com/ElMassimo/vite_ruby/commit/824b4ef8397828423d2ddda117bf27e365954961.

PR #492, opened 2024-09-04
fix: actually run vite inside bun

pigoz:
### Description 📖

Fixes the runner to actually use Bun and not Node

### Background 📜

When using bun, Runner would call into `bunx vite`, but the vite executable has a node shebang which bunx respects so vite would actually run inside node.

https://bun.sh/docs/cli/bunx#shebangs

### The Fix 🔨

This commit adds a --bun flag su that bunx ignores the node shebang.


ElMassimo (approved):

yaroslav:
FWIW as of now, even the `--bun` does not work on macOS. Bun still spawns Node when running vite. Hopefully, a temporary issue.

https://github.com/oven-sh/bun/issues/11869

PR #502, opened 2024-10-11
Add `logger` as a runtime dependency rather than default gem

joeldrapper:
### Description 📖

The default gem `logger` will be removed from Ruby 3.5.0 and must now be installed separately. This also resolves a warning in Ruby 3.3.5.

### Background 📜

Ruby is removing some of the “default” gems so they now need to be installed separately.

### The Fix 🔨

I’ve added the `logger` gem as a runtime dependency.

### Screenshots 📷

I don’t have any screenshots, but I have a copy of the warning message in Ruby 3.3.5.

> warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.

### DIFF
--- a/vite_ruby/vite_ruby.gemspec
+++ b/vite_ruby/vite_ruby.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'dry-cli', '>= 0.7', '< 2'
   s.add_dependency 'rack-proxy', '~> 0.6', '>= 0.6.1'
   s.add_dependency 'zeitwerk', '~> 2.2'
+  s.add_dependency 'logger', '~> 1.6'
 
   s.add_development_dependency 'm', '~> 1.5'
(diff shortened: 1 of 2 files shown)


vite_ruby/vite_ruby.gemspec
```diff
@@ -21,6 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'dry-cli', '>= 0.7', '< 2'
   s.add_dependency 'rack-proxy', '~> 0.6', '>= 0.6.1'
   s.add_dependency 'zeitwerk', '~> 2.2'
+  s.add_dependency 'logger', '~> 1.6'
```
joeldrapper, line 24: I don’t know if `vite_ruby` actually depends on `~> 1.6` or if it would be happy with an older version of the gem. It looks like `1.6.1` does support Ruby versions back to 2.5.0 so there shouldn’t be any issues there.

ElMassimo:
Thanks!

PR #506, opened 2024-10-21
Update vite_rails.gemspec to support Rails 8

qichunren:
### Description 📖

This pull request

### Background 📜

This was happening because

### The Fix 🔨

By changing


ElMassimo:
Addressed in [this commit](https://github.com/ElMassimo/vite_ruby/commit/563f5c78faa6d2df35e02c96df916c64994bed97).

I'd like to avoid open-ended requirements. A breaking change in `railties` for the APIs this gem uses seems unlikely, but it's possible.

PR #511, opened 2024-10-30
Update rubocop to version 1.28.2

mjankowski:
Bumps to the latest versions that still support ruby 2.5 as `TargetRubyVersion`, and regenerates todo file.

Related -- thoughts on bumping the minium ruby version here? Everything 3.0 and earlier is EOL at this point, and 3.1 will be at end of year. There are definitely some small language niceties to be found by bumping the rubocop target version and just autocorrecting...


mjankowski:
I had added the version limits because anything higher does not support the target ruby version in place at the time.

That said, I like this even better - https://github.com/ElMassimo/vite_ruby/pull/512 - so all good.

PR #540, opened 2025-04-10
Load .env environment variables for `vite dev`

ryansch:
### Description 📖

This pull request ensures that Rails has initialized before running the `vite dev` server.

### Background 📜

I noticed that setting `VITE_RUBY_*` environment variables via `.env.*` files would work for `vite build` but not for `vite dev`.

### The Fix 🔨

I generalized the `ViteRails::CLI::Build` module to apply to both `vite build` and `vite dev`.

### DIFF
--- a/vite_rails/lib/vite_rails/cli.rb
+++ b/vite_rails/lib/vite_rails/cli.rb
@@ -5,7 +5,7 @@ require "vite_rails"
 module ViteRails::CLI
 end
 
-module ViteRails::CLI::Build
+module ViteRails::CLI::EnsureRailsInit
   def call(**options)
     ensure_rails_init
     super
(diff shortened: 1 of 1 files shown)


ryansch:
We've been running this in production (and development) for almost a year without issues.

ryansch:
The problem I have is that without this patch, the server starts with an incorrect configuration.

I'm setting these 3: `VITE_RUBY_HOST`, `VITE_HMR_HOST`, and `VITE_HMR_CLIENT_PORT`. I'm using puma-dev for local development and this gets me full TLS with a proper hostname. It works great if `vite dev` loads me `.env.local` file.

ElMassimo:
Something that might achieve the same outcome while keeping the load time fast would be to detect if `Dotenv::Rails` is defined and load env files.

PR #542, opened 2025-04-13
fix: Check if `web` entry already on Profile before appending it

Andy9822:
### Description 📖

This pull request adds a logic to check wether the `Procfile.dev` already has a `web` entry. In case it does, we don't append a duplicate `web `entry.

### Background 📜

This was happening because the CLI code would append the entry without checking the file.
However, on Rails apps generated with the Rails generator, the `web` entry would already be in the Procfile. This would case 2 `web` entries on the same file

### The Fix 🔨

By changing the `append` method to be `append_if_no_web_entry` we can safely perform the action since we are now checking the file content.


ElMassimo:
Hi Andy!

I think it's a nice improvement, I'll test it when I have some time.

camertron:
I just ran into this as well and would love to see this PR get merged 😄

PR #549, opened 2025-05-20
docs: remove offline heroku links

iaurg:
### Description 📖

Remove offline demo apps and change heroku link on start to an anchor

### Background 📜

This was happening because all links to heroku are oggline

### The Fix 🔨

By changing links and removing old app links


ElMassimo:
Thanks!

PR #604, opened 2026-04-23
fix: Allow new Vite v8 `outputOptions` callback shape

raybrownco:
### Description 📖

Fix a build issue related to Vite 8's handling of the `assetFileNames` callback, which is manually typed in the `outputFileName()` function.

### Background 📜

Vite v8 (which uses Rolldown) changed the object shape passed to the `assetFileNames` callback.

In Vite v7 (which uses Rollup), the callback received a `PreRenderedAsset` object with a singular `name: string | undefined` field.

In Vite v8, Vite's internal `vite:css-post` plugin calls the same `assetFileNames` function with a plural `names: string[]` field instead. This matches the Rolldown `PreRenderedAsset` interface. See: https://github.com/rolldown/rolldown/blob/a71934bf/packages/rolldown/src/options/output-options.ts#L66-L77

### The Fix 🔨

First, we define a `PreRenderedAsset` interface that handles both shapes:

```ts
export interface PreRenderedAsset {
  name?: string // Vite v7, deprecated
  names?: string[] // Vite v8
  // ...
}
```

Then, determine which value to use within the `outputFileName()` function:

```ts
const outputFileName = (ext: string) => (asset: PreRenderedAsset) => {
    // Vite v8 uses `names`, earlier versions use `name`
    const resolvedName = asset.names?.[0] ?? asset.name ?? '[name]'
    const shortName = basename(resolvedName).split('.')[0]
    return posix.join(assetsDir, `${shortName}-[hash].${ext}`)
  }
```

### Screenshots 📷

A Rails project with `vite_rails` (3.0.19) and `vite_ruby` (3.9.2) began to fail after updating `vite` from 7.3.2 to 8.0.10:

Screenshot 2026-04-23 at 11 42 10 AM

### DIFF
--- a/vite-plugin-ruby/src/index.ts
+++ b/vite-plugin-ruby/src/index.ts
@@ -9,6 +9,15 @@ import { assetsManifestPlugin } from './manifest'
 
 export * from './types'
 
+export interface PreRenderedAsset {
+  type: 'asset'
+  name?: string // Vite v7, deprecated
+  names?: string[] // Vite v8
+  originalFileName?: string // Vite v7, deprecated
+  originalFileNames?: string[] // Vite v8
+  source: string | Uint8Array
+}
+
 // Public: The resolved project root.
 export const projectRoot = configOptionFromEnv('root') || process.cwd()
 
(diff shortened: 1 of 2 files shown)


raybrownco:
@ElMassimo Ready for your review: this PR fixes a bug that we ran into when attempting to upgrade Vite from v7 to v8 in one of our Rails projects. Thanks in advance for your consideration! Let me know if you object to any of these changes - I'm more than happy to accept feedback.

aTei:
Vite 8 needed for security fix of vite 7

raybrownco:
@ElMassimo Just following up on this. I won't ping you again out of respect for your capacity, but wanted to make sure that you saw this, in the case that you would _want_ to see it and just haven't seen it yet. Thanks for all that you do!

ElMassimo:
Hi Ray, looks great. Thanks for providing a detailed description!

Issue #605, opened 2026-04-24
Support bundledDev mode under Vite 8


scpike:
**Is your feature request related to a problem? Please describe.**
Vite 8 brings bundledDev (https://vite.dev/blog/announcing-vite8). vite_rails does not support this out of the box.

**Describe the solution you'd like**
I'd like to be able to enable bundledDev: true and have it *just work*. 

**Describe alternatives you've considered**

**Additional context**
bundledDev provides a very significant performance boost for my Rails app. This is a large frontend project (1m+ LOC) with hundreds of entrypoints and thousands of modules that get loaded on each page. In bundledDev mode it's very performant, but on regular mode (bypassing the Rails proxy, without which it's totally unusable) it can take tens of seconds to load a page.

Unfortunately it's not really straightforward to support bundledDev, and I understand you may want to wait until the feature settles on the vite side before trying to support it here. 

The fundamental thing is that bundledDev mode is sort of a hybrid between a pre-compiled mode and the regular vite dev server, and in bundledDev mode you don't get a `.vite/manifest.json` rendered from the vite dev server anymore. Vite itself in bundledDev mode has some middleware to figure out what to inject into <script> tags when it's managing index.html, but that is bypassed by vite_rails.

You can work around this with custom middleware on the vite side that emits a manifest anyway, and then something custom on the ruby side to pull that in. I had claude write me a POC of this that's working for me for my app, but I don't really feel comfortable suggesting it as a PR. I may try to wrap my head around this more and put up a proper PR if I can, but frankly someone with more vite_rails knowledge could probably come up with a better solution.

andriytyurnikov:
#607 is a prerequisite — it lifts `vite-plugin-ruby`'s peer-dep to `vite >=8` and confirms the workspace plugins build cleanly under Rolldown. Doesn't wire `bundledDev` itself, but unblocks that work.

andriytyurnikov:
Looked at what supporting `bundledDev` actually entails. Three surface areas:

1. **Manifest emission during dev.** `vite-plugin-ruby`'s `assetsManifestPlugin` writes the manifest at build close. In bundledDev, Vite holds bundle output in `MemoryFiles` with `watch: { skipWrite: true }` (see `FullBundleDevEnvironment` in `vite/dist/node/chunks/node.js`). The plugin would need to emit the manifest in-memory and serve it via custom middleware.

[...]

Issue #592, opened 2026-03-10
Support Vite 7 - Failed compatibility check


Splines:
When trying to upgrade our dependencies at [MaMpf](https://github.com/mampf-hd/mampf), we faced [this error](https://github.com/MaMpf-HD/mampf/actions/runs/22916758983/job/66504134658?pr=1068#step:5:996):

```
ERROR -- : Failed to check mode for Vite: vite-plugin-ruby@^5.1.0 might not be compatible with vite_ruby-3.9.3

You may disable this check if needed: https://vite-ruby.netlify.app/config/#skipcompatibilitycheck

You may upgrade both by running:

     bundle exec vite upgrade

bin/rails aborted!
```

This error occurs with:

- Gem: `vite_ruby: 3.9.3`
- Npm package: `vite-plugin-ruby: 5.1.3`

`bundle exec vite upgrade` changes the `package.json` from `vite@^7.0.0` to `vite@^6.2.6`, resolving to version `6.4.1`, which was [released](https://github.com/vitejs/vite/releases/tag/v6.4.1) October 20, 2025. Given that Vite 8 is already in Beta, I'd love to see support for Vite 7 for the awesome vite_ruby😊 We were already using Vite 7 in production, but just faced this issue now during the update.

andriytyurnikov:
FYI #607 takes the path to Vite 8 directly (skipping Vite 7) and aligns `vite-plugin-ruby ^5.2` with `DEFAULT_VITE_VERSION = "^8.0.0"`, which `main` already pins. Once that lands, the compat-check failure you hit on Vite 7 is moot — `vite upgrade` from Vite 5/6 → Vite 8 with the new plugin range no longer trips the warning.

ElMassimo:
Hi! Thanks for reporting.

Vite 8 support was released in `vite-plugin-ruby@5.2.0` and `vite_ruby@3.10.0`, so upgrading both packages resolves the compatibility check without downgrading Vite.

Closing as resolved.

Issue #527, opened 2025-01-18
The requested module 'vite' does not provide an export named 'defaultClientConditions'


socketopp:
[troubleshooting section]: https://vite-ruby.netlify.app/guide/troubleshooting.html

- [x] I have tried upgrading by running `bundle update vite_ruby`.
- [x] I have read the __[troubleshooting section]__ before opening an issue.

### Description 📖

I was following these instructions to create a rails project with svelte using vite_rails. 
But when running `bin/vite dev`

```
failed to load config from /ruby/my_app/vite.config.ts
error when starting dev server:
file:/ruby/my_app/node_modules/@sveltejs/vite-plugin-svelte/src/utils/options.js:5
        defaultClientConditions,
        ^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module 'vite' does not provide an export named 'defaultClientConditions'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:171:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:254:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:475:24)
    at async loadConfigFromBundledFile (file:/ruby/my_app/node_modules/vite/dist/node/chunks/dep-CB_7IfJ-.js:66691:15)
    at async loadConfigFromFile (file:/ruby/my_app/node_modules/vite/dist/node/chunks/dep-CB_7IfJ-.js:66532:24)
    at async resolveConfig (file:/ruby/my_app/node_modules/vite/dist/node/chunks/dep-CB_7IfJ-.js:66140:24)
    at async _createServer (file:/ruby/my_app/node_modules/vite/dist/node/chunks/dep-CB_7IfJ-.js:62758:18)
    at async CAC.<anonymous> (file:/ruby/my_app/node_modules/vite/dist/node/cli.js:735:20)
FAIL
``` 

Instructions:
```
Create a Rails Application
Run the following command to create a new Rails app:

rails new my_app --skip-javascript

The --skip-javascript flag is optional, but it avoids setting up the default Rails JavaScript configuration since you'll be using Vite.

Add the vite_rails Gem
Add the vite_rails gem to your Gemfile:

gem "vite_rails"

Then run:

bundle install

Install Vite
Run the Vite installer provided by the gem:

bin/vite install

Install Svelte
Install Svelte as a dependency using npm or Yarn:

yarn add svelte

Also, install the necessary Svelte integrations for Vite:

[...]

PR #607, opened 2026-04-27
Vite 8, Rails 8

### LINKED ISSUE #588: Sourcemaps enabled in production by default (security issue)
[troubleshooting section]: https://vite-ruby.netlify.app/guide/troubleshooting.html

- [x] I have tried upgrading by running `bundle update vite_ruby`.
- [x] I have read the __[troubleshooting section]__ before opening an issue.

### Description 

**vite-plugin-ruby has a backwards sourcemap default that enables sourcemaps in production and disables them in development.** This is a security vulnerability that exposes application source code in production builds.

The plugin sets `build.sourcemap: !isLocal` in `src/index.ts` (~line 155), which results in:
- **Production mode** (`!isLocal` = `true`): ✗ Sourcemaps ENABLED - exposes source code
- **Development mode** (`!isLocal` = `false`): ✗ Sourcemaps DISABLED - breaks debugging
- **Test mode** (`!isLocal` = `false`): ✗ Sourcemaps DISABLED

This is the **opposite** of Vite's secure default (`build.sourcemap: false`).

**Security Impact:**
- `.js.map` and `.css.map` files publicly accessible in production
- `sourceMappingURL` comments present in minified bundles
- Complete unminified source code visible in browser DevTools
- Business logic, API endpoints, and implementation details exposed

**Root Cause:**

In `vite-plugin-ruby/src/index.ts`:
```typescript
const isLocal = config.mode === "development" || config.mode === "test";
const build = {
  emptyOutDir: ...,
  sourcemap: !isLocal,  // ← BUG: true in production, false in dev!
  ...userConfig.build,
  // ...
}
```

The logic appears inverted - `!isLocal` enables sourcemaps when NOT in local development.

### Reproduction 

**Minimal reproduction:**

1. Create a fresh Rails app with vite_ruby (default config)
2. Use default `vite.config.ts` with NO explicit `build.sourcemap` setting:

```typescript
// vite.config.ts - using plugin defaults
import { defineConfig } from 'vite'
import RubyPlugin from 'vite-plugin-ruby'

export default defineConfig({
  plugins: [RubyPlugin()],
  build: {
    // NO sourcemap config - inherits plugin default
  }
})
```

3. Build for production:
```bash
rm -rf public/vite
RAILS_ENV=production NODE_ENV=production bin/vite build
```

4. Check for sourcemap exposure:
```bash
find public/vite/assets -name "*.js.map"
grep -r "sourceMappingURL" public/vite/assets/*.js
```

**Expected:** No sourcemap files, no sourceMappingURL comments
**Actual:** Sourcemap files present, comments pointing to them

5. Check development build:
```bash
rm -rf public/vite-dev
RAILS_ENV=development NODE_ENV=development bin/vite build
find public/vite-dev/assets -name "*.js.map"
```

**Expected:** Sourcemap files present for debugging
**Actual:** No sourcemap files (debugging broken)

<details>
<summary>Vite Ruby Info</summary>

```
bin/vite present?: true
vite_ruby: 3.9.1
vite_rails: 3.0.19
rails: 7.2.2.1
ruby: ruby 3.3.7 (2025-01-15 revision be31f993d7) [arm64-darwin24]
node: v20.12.2
yarn: 4.1.1

installed packages:
metarina@0.1.0
├─┬ vite-plugin-ruby@5.1.1
│ └── vite@5.4.21 deduped
└── vite@5.4.21
```
</details>

## Proposed Fix

**Option 1: Remove the default (recommended)**

Let Vite's default apply (sourcemap: false):

```typescript
const isLocal = config.mode === "development" || config.mode === "test";
const build = {
  emptyOutDir: ...,
  // REMOVE: sourcemap: !isLocal,
  ...userConfig.build,  // User can still override
  // ...
}
```

**Option 2: Invert the logic**

Enable sourcemaps only in development:

```typescript
const isLocal = config.mode === "development" || config.mode === "test";
const build = {
  emptyOutDir: ...,
  sourcemap: config.mode === 'development',  // Only in dev
  ...userConfig.build,
  // ...
}
```

## Current Workaround

Users must explicitly override in `vite.config.ts`:

```typescript
export default defineConfig(({ mode }) => ({
  build: {
    sourcemap: mode === 'development'  // Override plugin default
  }
}))
```

## Impact

This affects **all vite_ruby users** who don't explicitly set `build.sourcemap` in their config. Production applications unknowingly expose their complete source code.

## References

- **Vite docs**: https://vitejs.dev/config/build-options.html#build-sourcemap (default: `false`)
- **vite-plugin-ruby source**: https://github.com/ElMassimo/vite_ruby/blob/main/vite-plugin-ruby/src/index.ts#L155

---

Happy to submit a PR if you'd like! This appears to be an unintentional logic inversion.

andriytyurnikov:
## Summary

Brings vite_ruby and the workspace plugins into compatibility with Vite 8 (Rolldown-powered) and refreshes the CI matrix and Ruby framework targets. Also fixes two latent runtime bugs, a sourcemap-in-production security default, and a test-determinism issue surfaced during the upgrade.

### Vite 8 compatibility
- `vite-plugin-ruby` and `vite-plugin-rails` peer-dep `vite >=8.0.0`.
- `examples/rails`: gate the `@administrator/` alias and `server.fs.allow` entries on `ADMINISTRATOR_ASSETS_PATH` actually being set. The previous code used a TypeScript non-null assertion (`process.env.X!`), so when the env var was unset, `server.fs.allow` received `[undefined]`. Vite ≤7 silently dropped non-string entries; Vite 8 strict-validates them and crashes with `TypeError [ERR_INVALID_ARG_TYPE]: paths[1] must be of type string`.

### Bug fixes
- **Closes #588** — `vite-plugin-ruby`: stop forcing `build.sourcemap = !isLocal`. The inverted default was publishing `.js.map` files in production (exposing application source) and disabling them in development (breaking debugging). Drop the override; Vite's safe default of `false` now applies in every mode, and consumers opt in via their own `build.sourcemap` config when they actually want sourcemaps.
- `vite-plugin-ruby` dev middleware: `return` after `res.end()` in the `/index.html` 404 fallback so downstream middleware doesn't write to a closed response.
- `vite_ruby`: synchronize `dev_server_running?` cache with a `Mutex`. The prior read of `@running` could race with the write of `@running_checked_at` and raise `NoMethodError` on a cold concurrent call from a multi-threaded server (Puma).

### Test determinism
- Fix `vite-plugin-rails/example`'s build script to pin `NODE_ENV=production`. Without it, vitest's default `NODE_ENV=test` leaked into the test-driven build, shipping Vue's *development* bundle (with `Object.freeze`, prop validators, and `@vitejs/plugin-vue`'s absolute-path `__file` injection — ~27 KB extra and per-machine paths).
- `build.spec.ts`: strip 8-char content hashes and replace SRI values with a `<integrity>` placeholder so the snapshot survives routine dep bumps that change chunk bytes without changing manifest shape.

### CI / tooling
- Expand the Ruby matrix to 3.3 / 3.4 (gemspec floor) and Rails 7.0–8.1 + edge.
- Drop Ruby 3.1 / 3.2 (below gemspec floor — `bundle install` would fail).
- Bump `setup-node` to 22 (Vite 8 requires Node 20.19+/22.12+).
- JS workflow: test both `vite-plugin-ruby` and `vite-plugin-rails`, build all workspace packages, install at workspace root.
- Enable `corepack` so the test_app's yarn-driven setup finds `yarn`.
- Regenerate `test/test_app/yarn.lock` for the Vite 8 bump.

### Related (likely resolved as side effects)
- #527 — `'vite' does not provide an export named 'defaultClientConditions'`. Tightening the peer-dep to `vite >=8` and pinning `DEFAULT_VITE_VERSION = "^8.0.0"` sidesteps the mixed-version state that triggered the import error.
- #592 — "Support Vite 7 — Failed compatibility check". The compat-check pain reported there is removed once consumers move to Vite 8 with `vite-plugin-ruby ^5.2.0` (which `DEFAULT_PLUGIN_VERSION` already targets on `main`).
- #605 — "Support `bundledDev` mode under Vite 8". This PR doesn't wire bundledDev itself but is the prerequisite for that work.

## Test plan

- [x] `pnpm -r build` — both workspace packages build clean.
- [x] `pnpm -C vite-plugin-ruby test --run` and `pnpm -C vite-plugin-rails test --run` pass.
- [x] Local `examples/rails` `vite build` succeeds with `ADMINISTRATOR_ASSETS_PATH` pointing at `example_engine/app/frontend`.
- [ ] Confirm the Ruby matrix passes in your CI.


ElMassimo:
Hi Andriy!

I'd appreciate it if you would break these changes into smaller PRs, otherwise I can't accept your contribution.

andriytyurnikov:
Hello @ElMassimo, apologies for noise.
Please look at this PR as a Proof of Concept - informative-first.

That being said - I would like to highlight, that 90% of changes are changes in `generated` files, pnpn lock, etc - and fundamentally those have be changed one way or another.

I see what I can do to split it though.

andriytyurnikov:
@ElMassimo first piece is published https://github.com/ElMassimo/vite_ruby/pull/608

PR #610, opened 2026-07-03
fix: stop overriding Rails' media default in vite_javascript_tag

### LINKED ISSUE #535: Remove default media: 'screen' for stylesheet tag in line with Rails defaults
Rails now doesn't include the media attribute by default for `stylesheet_link_tag` but the default for `vite_javascript_tag` has `media: 'screen'`:

https://github.com/ElMassimo/vite_ruby/blob/98d3d5f452d82cfc25720ed85be83f4b43b45d93/vite_rails/lib/vite_rails/tag_helpers.rb#L44

https://github.com/rails/rails/issues/41213

It'd be a breaking change but would keep things in line with the new rails defaults.

`vite_stylesheet_tag` would now also not return a `media` attribute since it relies on `stylesheet_link_tag` without any default options.

ErnaneJ:
### Description 📖
Removes the hardcoded `media: "screen"` default from `vite_javascript_tag` so it defers to Rails' own `apply_stylesheet_media_default` setting instead of overriding it.

Fixes #535

### Background 📜
Rails introduced `config.action_view.apply_stylesheet_media_default` to let apps opt out of the legacy `media="screen"` behavior. Because `vite_javascript_tag` was explicitly passing `media: "screen"` to `stylesheet_link_tag`, though, this config had no effect on stylesheet tags generated by Vite entrypoints. Users who set `apply_stylesheet_media_default = false` were still getting `media="screen"`.

### The Fix 🔨
`vite_javascript_tag` now passes `media: nil` instead of `media: "screen"`, so `stylesheet_link_tag` controls the attribute. When `apply_stylesheet_media_default` is `true` (the Rails default), behavior is unchanged. When it's `false`, no `media` attribute is rendered. Callers can still pass an explicit `media:` value if needed.

The change covers both `vite_rails` and `vite_rails_legacy`.

### DIFF
--- a/vite_rails/lib/vite_rails/tag_helpers.rb
+++ b/vite_rails/lib/vite_rails/tag_helpers.rb
@@ -41,7 +41,7 @@ module ViteRails::TagHelpers
     skip_preload_tags: false,
     skip_style_tags: false,
     crossorigin: "",
-    media: "screen",
+    media: nil,
     **options)
     entries = vite_manifest.resolve_entries(*names, type: asset_type)
     tags = javascript_include_tag(*entries.fetch(:scripts), crossorigin: crossorigin, type: type, extname: false, **options)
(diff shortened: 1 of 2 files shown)


ElMassimo:
Hi Ernane! Given the setting was introduced in Rails 7, I think it makes sense to update `vite_rails`, but not `vite_rails_legacy`.

Thanks!

brendon:
Thanks for this! :D

PR #617, opened 2026-08-10
fix: allow secure rack-proxy releases

Ahsan-Hoque:
## Summary

Relax the `rack-proxy` runtime dependency from `~> 0.6` to `>= 0.6.1, < 2`. This keeps existing 0.x versions supported and allows applications to resolve the secure `rack-proxy` 1.x releases, including 1.0.1.

## Validation

- `bundle install`
- `bundle exec ruby -Itest test/dev_server_proxy_test.rb`
- `bundle exec ruby -Itest test/compatibility_check_test.rb test/config_test.rb`
- `gem build vite_ruby.gemspec`

The full `bundle exec rake test` run reached the existing mounted-app dependency-update integration test and exceeded the local timeout without reporting a product test failure.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -18,7 +18,7 @@ PATH
       dry-cli (>= 0.7, < 2)
       logger (~> 1.6)
       mutex_m
-      rack-proxy (~> 0.6, >= 0.6.1)
+      rack-proxy (>= 0.6.1, < 2)
       zeitwerk (~> 2.2)
 
 GEM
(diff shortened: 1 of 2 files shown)


ElMassimo:
Hi Ahsan!

I've relaxed the requirement for `rack-proxy` in https://github.com/ElMassimo/vite_ruby/commit/37b1edb5ee74704192ed0b23efb6c4976e120e7d.

I don't think an upper constraint is [desirable](https://byroot.github.io/ruby/bundler/2026/04/20/bundle-features.html#upper-constraints).

PR #620, opened 2026-08-27
fix: use server.ws.clientPort on Vite 8.1+

### LINKED ISSUE #614: vite-plugin-ruby 5.2.2 triggers deprecation warning under vite 8.1.5
```
$ ./bin/vite dev
Trace: `server.hmr.protocol/host/port/path/clientPort/timeout/server` is deprecated. Use `server.ws.*` instead. Note that this option may be set by a plugin.
```

Triggered by:

https://github.com/ElMassimo/vite_ruby/blob/3c59d7bf969b4d34a4db9ed890bc1f12aed56da0/vite-plugin-ruby/src/config.ts#L96-L100

And possibly subsequent code.

iuhoay:
### Description 📖

vite-plugin-ruby still writes `server.hmr.clientPort`, which Vite 8.1 logs as deprecated on every `bin/vite dev`.

Closes #614.

### Background 📜

rack-proxy does not forward websockets, so the plugin sets `clientPort` to the Vite port. Vite 8.1 moved that option to `server.ws.*`; Vite 8.0 still reads `server.hmr.clientPort`.

### The Fix 🔨

Write `server.ws.clientPort` from Vite 8.1 onward, and keep the old assignment for Vite 5–8.0.

### Screenshots 📷

### DIFF
--- a/vite-plugin-ruby/src/config.ts
+++ b/vite-plugin-ruby/src/config.ts
@@ -1,9 +1,10 @@
 import { join, relative, resolve } from 'path'
 import { globSync } from 'tinyglobby'
 
+import { version as viteVersion } from 'vite'
 import type { UserConfig, ServerOptions } from 'vite'
 import { APP_ENV, ALL_ENVS_KEY, ENTRYPOINT_TYPES_REGEX } from './constants'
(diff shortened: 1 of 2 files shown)


ElMassimo:
Looks great, thanks!

PR #625, opened 2026-09-16
perf: skip re-hashing unchanged watched files on every build check

### LINKED ISSUE #624: autoBuild: the staleness check re-hashes every watched file on each helper call
### Description

With `autoBuild` enabled, every call to a `vite_*` helper can re-read and re-hash every watched file. On an app with a large frontend this dominates the cost of rendering a layout, and in a test suite it dwarfs the tests themselves.

Below are measurements and a fix that keeps `autoBuild` behaving exactly as it does today. I have it implemented with tests against the gem's own suite, and I'd be glad to send a PR if you like the approach.

### Where the time goes

`Manifest#lookup` calls `builder.build` on every lookup, which calls `watched_files_digest`, which globs `config.watched_paths` and runs `Digest::SHA1.file` over each file. `@last_digest_at` memoizes the result, but only for one second.

Our app has ~1850 watched files under `app/javascript`, 72 MB in total. Cost of one digest pass, measured:

| | `Dir[]` | `File.stat` × 1846 | `Digest::SHA1.file` × 1846 | total |
|---|---|---|---|---|
| Linux, native filesystem | 2.1 ms | 1.9 ms | **104.7 ms** | ~100 ms |
| Linux container, macOS bind mount | 56.3 ms | 3.5 ms | **364.5 ms** | ~460 ms |

The traversal is cheap. Reading 72 MB off disk to hash it is the whole cost, and the answer is nearly always "not stale".

The one-second memo decides how often that price is paid, and it works against you exactly when things are slow: the slower the render, the more likely the next helper call falls outside the window. Our layout makes five lookups that reach the manifest. Locally they run back to back and coalesce into one pass; on our CI runner they do not — a captured log from one run shows 540 `build` calls, i.e. roughly one pass per lookup rather than one per render.

Instrumenting a single integration-test file (29 tests, Inertia layout), `autoBuild: true`:

| | `build` calls | digest passes | time inside the staleness check |
|---|---|---|---|
| v3.10.5 | 64 | 5 | **2052 ms** |
| with the fix below | 64 | 4 | **232 ms** |

### Proposed fix: cache per-file digests, keyed on mtime and size

Keep hashing contents, but read a file again only once it has been touched:

```ruby
def watched_files_digest
  return @last_digest if @last_digest_at && Time.now - @last_digest_at < 1

  config.within_root do
    previous_digests, @file_digests = @file_digests, {}
    file_ids = Dir[*config.watched_paths].sort.map { |file|
      file_digest(file, previous_digests)
    }.compact
    @last_digest_at = Time.now
    @last_digest = Digest::SHA1.hexdigest(file_ids.join("/"))
  end
end

def file_digest(file, previous_digests)
  stat = File.stat(file)
  return if stat.directory?

  signature = [stat.mtime, stat.size]
  previous_signature, previous_id = previous_digests[file]
  id = if previous_signature == signature && !recently_modified?(stat)
    previous_id
  else
    "#{File.basename(file)}/#{Digest::SHA1.file(file).hexdigest}"
  end
  @file_digests[file] = [signature, id]
  id
end

# A file edited twice within the resolution of the file system clock, without
# changing its size, would keep the same signature. Rehashing the files touched
# in the last second rules that out, the same way Git resolves a racily clean
# entry in the index.
def recently_modified?(stat)
  Time.now - stat.mtime < 1
end
```

Properties worth calling out:

- **The digest stays content-based.** A fresh checkout that changes mtimes but not contents yields the same digest and triggers no rebuild. That is why the cache is keyed on the stat rather than the digest being replaced by one.
- **No new class of missed edit.** mtime plus size can collide only for an edit landing within the clock resolution of an earlier one, and `recently_modified?` re-reads exactly those files — the rule Git applies to racily clean index entries.
- **`stale?` and the contents of `last-build-*.json` are unchanged.** The cache lives on the `Builder` instance and is rebuilt on every pass, so deleted files drop out of it on their own.

Measured on the same trees as above, over repeated passes with no file changes:

| | v3.10.5 | with the fix |
|---|---|---|
| Linux, native filesystem | 99.8 ms | **6.3 ms** |
| Linux container, macOS bind mount | ~460 ms | **~58 ms** |

### A second, smaller one

`Manifest#manifest` calls `refresh` on every lookup when `auto_build` is on, re-reading and re-parsing `manifest.json` and rebuilding `resolve_references` each time. Ours is 980 KB with 2202 entries, which measured at ~13–21 ms per lookup, or a little over a second across the 64 lookups in the run above. Memoizing on the manifest file's mtime and size would keep the "always fresh" behaviour while paying for it once per build.

Happy to split these into two PRs, or to send only the first.

### Why not just turn `autoBuild` off

We did, and it works, but a stale build then no longer rebuilds itself and everyone has to remember to build after touching the frontend. You made the case in #291 and #376 that `autoBuild` is an important part of the DX, and we agree — the check being expensive looks like the more fixable half of the problem.

### Versions

`vite_ruby` 3.10.5, `vite_rails` 3.11.1, Ruby 4.0, Rails 8.

mokevnin:
### Description 📖

This pull request makes the `autoBuild` staleness check stop re-reading watched files that have not changed. Closes #624, where the measurements are written up in full.

### Background 📜

Every call to a `vite_*` helper goes through `Manifest#lookup`, which calls `builder.build`, which calls `watched_files_digest` — a glob over `config.watched_paths` plus `Digest::SHA1.file` on every one of them. `@last_digest_at` memoizes the result, but only for a second.

On an app with ~1850 watched files (72 MB under `app/javascript`) one pass costs:

| | `Dir[]` | `File.stat` × 1846 | `Digest::SHA1.file` × 1846 |
|---|---|---|---|
| Linux, native filesystem | 2.1 ms | 1.9 ms | **104.7 ms** |
| Linux container, macOS bind mount | 56.3 ms | 3.5 ms | **364.5 ms** |

The traversal is cheap; reading 72 MB to hash it is the whole cost, and the answer is nearly always "not stale".

The one-second memo decides how often that is paid, and it works against you exactly when things are slow: the slower the render, the more likely the next helper call falls outside the window. Our layout makes five lookups that reach the manifest. Locally they coalesce into one pass; on our CI runner they do not — a captured log shows 540 `build` calls in one run.

### The Fix 🔨

By keeping each file's id under its `mtime` and size, and reading the file again only once that changes.

| repeated pass, no file changes | before | after |
|---|---|---|
| Linux, native filesystem | 99.8 ms | **6.3 ms** |
| Linux container, macOS bind mount | ~460 ms | **~58 ms** |

In one integration-test file (29 tests, `autoBuild: true`) the time spent inside the staleness check went from 2052 ms to 232 ms, over the same 64 `build` calls.

Three things worth calling out:

- **The digest stays content-based.** A fresh checkout that changes mtimes but not contents produces the same digest and triggers no build. That is why the cache is keyed on the stat instead of the digest itself being replaced by one.
- **No new class of missed edit.** `mtime` plus size can collide only for an edit landing within the clock resolution of an earlier one, and `recently_modified?` re-reads exactly those files — the rule Git applies to racily clean index entries.
- **`Build#stale?` and the contents of `last-build-*.json` are untouched.** The cache lives on the `Builder` instance and is rebuilt on each pass, so deleted files drop out of it on their own.

Two tests are added: one asserts that a second pass reads nothing from disk and that touching a file re-reads that file and only that file; the other asserts that changed contents still change the digest.

`#624` also names a second, smaller cost — `Manifest#manifest` re-reading and re-parsing `manifest.json` on every lookup when `auto_build` is on. That one is not in this pull request; happy to send it separately if you want it.


ElMassimo:
Hi Kirill! 

I like this optimization, it's small and self-contained, so it feels like a reasonable addition.

That said, I notice that you mention `auto_build` being triggered in the CI, which seems wasteful.

[As mentioned in the guide](https://vite-ruby.netlify.app/guide/development.html#integration-tests-in-the-ci-%E2%9C%85), it's recommended to set `VITE_RUBY_AUTO_BUILD="false"` in your CI to prevent the check from ever running.

mokevnin:
Thanks for merging, and for the pointer on `VITE_RUBY_AUTO_BUILD` — agreed, we've since turned auto-build off in the test environment, CI included.

